PR #23, opened 2014-04-17
Adding support for Consul tests

armon:
This is probably a very newb PR so please forgive. I've cargo-culted most of this to make it work based on the etcd work. The one thing is it seems Jepsen does value based CAS while Consul does index based CAS. I've worked around to first do a read to get the index and then do the CAS operation. Not sure if there is a way to improve this by passing the index from a read into the CAS operation. Thanks!

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -25,6 +25,7 @@
 ;                            :exclusions [com.datastax.cassandra/cassandra-driver-core]]
 ;                           [com.datastax.cassandra/cassandra-driver-core "1.0.3"]
                            [byte-streams "0.1.4"]
+                           [base64-clj "0.1.1"]
                            [com.novemberain/langohr "2.7.1"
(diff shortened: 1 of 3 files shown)


aphyr:
OK, all this looks generally good to me! Sorry about the insane wait, I'm whittling down the PR backlog this month. :)

jpfuentes2:
Sounds good to me. Do you want me to try my hand at resolving merge conflicts?

aphyr:
Got it merged, porting your stuff forward to the new code structure now.

aphyr:
OK, so I hooked up the dependency machinery and all; you can run the tests with `lein with-profile +consul test jepsen.system.consul-test`, but they crash with `/var/log/consul.log: Permission denied`; probably just an issue in the setup permissions but I'm not gonna dig into it too deep right now.

PR #48, opened 2015-01-27
Support etcd v2.0

ahorn:
Newer versions of etcd have deprecated many essential command line
options for setting up a static cluster [1]. This caused exceptions
on the cluster machines and also in the Jepsen script. This patch
tweaks things so that the cluster will start up with etcd v2.0.

Note that the LXC instructions for setting up the host names are
slightly adjusted.

[1] https://github.com/coreos/etcd/blob/master/Documentation/clustering.md#04-to-20-migration-guide

### DIFF
--- a/etcd/src/jepsen/system/etcd.clj
+++ b/etcd/src/jepsen/system/etcd.clj
@@ -24,19 +24,19 @@
 (def log-file "/var/log/etcd.log")
 
 (defn peer-addr [node]
-  (str (net/ip node) ":2380"))
+  (str (name node) ":2380"))
 
 (defn addr [node]
-  (str (net/ip node) ":2380"))
+  (str (name node) ":2380"))
 
 (defn cluster-url [node]
-  (str "http://" (net/ip node) ":2380"))
+  (str "http://" (name node) ":2380"))
 
 (defn listen-client-url [node]
-  (str "http://" (net/ip node) ":2379"))
+  (str "http://" (name node) ":2379"))
 
 (defn cluster-info [node]
-  (str (name node) "=http://" (net/ip node) ":2380"))
+  (str (name node) "=http://" (name node) ":2380"))
 
 (defn peers
   "The command-line peer list for an etcd cluster."
--- a/etcd/src/jepsen/system/etcd.clj
+++ b/etcd/src/jepsen/system/etcd.clj
@@ -24,17 +24,26 @@
 (def log-file "/var/log/etcd.log")
 
 (defn peer-addr [node]
-  (str (name node) ":7001"))
+  (str (net/ip node) ":2380"))
 
 (defn addr [node]
-  (str (name node) ":4001"))
+  (str (net/ip node) ":2380"))
+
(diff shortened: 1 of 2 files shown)


etcd/src/jepsen/system/etcd.clj
```diff
@@ -24,17 +24,26 @@
 (def log-file "/var/log/etcd.log")
 
 (defn peer-addr [node]
-  (str (name node) ":7001"))
+  (str (net/ip node) ":2380"))
 
 (defn addr [node]
-  (str (name node) ":4001"))
+  (str (net/ip node) ":2380"))
```
aphyr: Does etcd no longer do name resolution? Should be pretty fast, they're hardcoded in /etc/hosts.

lxc.md
```diff
@@ -123,7 +123,6 @@ Set up hostfiles on each box with hardcoded IP addresses
 
 
 ```
-127.0.0.1 localhost n1 n1.local
```
aphyr: Heyyyy I'm concerned this is gonna break a ton of stuff. Why change it?

ahorn:
Hi there. I am not an expert with etcd but I agree that the current fix is probably not the most general setup. I'll try out another configuration and let you know what I find out.

lxc.md
ahorn: I've addressed the `net/ip` issue, so that's fixed. But the localhost issue remains: when `127.0.0.1 localhost n1 n1.local` is removed from /etc/hosts everything works fine, but when localhost is specified in etc/hosts, the cluster machines cannot talk to each other:
```
2015/01/30 11:48:24 raft: 5440ff22fe632778 [logterm: 1, index: 5] sent vote request to 9b116f88cab4dc9 at term 2
2015/01/30 11:48:24 sender: error posting to 5aa594b5d9b66c42: dial tcp 192.168.122.12:2380: connection refused
2015/01/30 11:48:24 sender: the connection with 5aa594b5d9b66c42 becomes inactive
```

Slightly strange.

aphyr: Ohhh hang on, I wonder if your notes aren't listening on loopback, and they're trying to initiate a connection back to their own hostname.

ahorn: I just followed the guide https://github.com/coreos/etcd/blob/master/Documentation/clustering.md#static where this doesn't seem to be an issue. I'll ping the etcd crowd and see if they can offer any advice.

ahorn: Yicheng Qin from etcd says [1] a possible reason may be that listen-peer-urls is at localhost:2380 by default, which cannot receive message from 192.168.122.12:2380.

[1] https://groups.google.com/d/msg/etcd-dev/AslsLq2Lddk/83L_3jBcfNUJ

aphyr:
OK, I'm gonna merge this and just remove the part about /etc/hosts for now.

aphyr:
Thanks for all your hard work, @ahorn! :)

ahorn:
Cheers @aphyr!

PR #51, opened 2015-04-07
Various cleanups for Elasticsearch test

dakrone:
This change makes four changes:

- Adding a logging configuration

In order to actually determine the root cause of issues, more verbose logging is needed. This defaults to more verbose logging for Elasticsearch and adds the ability to change it from Jepsen in the future (instead of manually by hand).

- Moving `nuke!` to *before* tests

Without this, Jepsen deletes all traces of itself after running, which makes debugging much more difficult (no logs and no data left).

- Use more reasonable settings for scroll

An optional change, but I figured I would make this change anyway.

- Wait for index to become green after creation

This is a key part of testing Elasticsearch, and clients should always do this when creating indices.

------------------------------------

Note that I was able to reproduce the failures in https://github.com/elastic/elasticsearch/issues/10426 (about half of the time) **without** these changes, however after the change which waits for green after index creation, I am no longer able to reproduce data loss with the `create-pause` test (still evaluating the other tests).

### DIFF
--- a/elasticsearch/src/elasticsearch/core.clj
+++ b/elasticsearch/src/elasticsearch/core.clj
@@ -173,6 +173,15 @@
 
 (def index-name "jepsen-index")
 
+(defn index-already-exists-error?
+  "Return true if the error is due to the index already existing, false
+  otherwise."
+  [error]
+  (and
+   (-> error .getData :status (= 400))
+   (re-find #"IndexAlreadyExistsException"
+            (http-error error))))
+
 (defrecord CreateSetClient [client]
   client/Client
   (setup! [_ test node]
@@ -185,11 +194,13 @@
                                          {:num {:type "integer"
                                                 :store "yes"}}}}
                     :settings {"index" {"refresh_interval" "1s"}})
-        (catch Throwable t))
+        (catch clojure.lang.ExceptionInfo e
+          (when-not (index-already-exists-error? e)
+            (throw e))))
       (esa/cluster-health client
                           {:index [index-name] :level "indices"
                            :wait_for_status "green"
-                           :wait_for_nodes "5"})
+                           :wait_for_nodes (count (:nodes test))})
       (CreateSetClient. client)))
 
   (invoke! [this test op]
@@ -247,11 +258,13 @@
       (try
         (esi/create client index-name
                     :mappings {mapping-type {:properties {}}})
-        (catch Throwable t))
+        (catch clojure.lang.ExceptionInfo e
+          (when-not (index-already-exists-error? e)
+            (throw e))))
       (esa/cluster-health client
                           {:index [index-name] :level "indices"
                            :wait_for_status "green"
-                           :wait_for_nodes "5"})
+                           :wait_for_nodes (count (:nodes test))})
 
       ; Initial empty set
       (esd/create client index-name mapping-type {:values []} :id doc-id)
--- a/elasticsearch/src/elasticsearch/core.clj
+++ b/elasticsearch/src/elasticsearch/core.clj
@@ -20,6 +20,7 @@
             [jepsen.os.debian         :as debian]
             [clj-http.client          :as http]
             [clojurewerkz.elastisch.rest          :as es]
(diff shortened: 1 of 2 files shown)


dakrone:
@aphyr also, how would you feel about me replacing Elastisch with vanilla clj-http? Elastisch uses clj-http internally anyway and it would reduce the number of moving parts in this test. I'm happy to submit another PR if you are interested.

elasticsearch/src/elasticsearch/core.clj
```diff
@@ -175,20 +178,18 @@
...
-                    :settings {"index" {"refresh_interval" "1"}})
-        (catch clojure.lang.ExceptionInfo e
-          ; Is this seriously how you're supposed to do idempotent
-          ; index creation? I've gotta be doing this wrong.
-          (let [err (http-error e)]
-            (when-not (re-find #"IndexAlreadyExistsException" err)
-              (throw (RuntimeException. err))))))
-
+                    :settings {"index" {"refresh_interval" "1s"}})
+        (catch Throwable t))
```
aphyr: I'd really prefer to know about connection errors etc that happen here; the only reason it's appropriate to noop is if the index already exists.

elasticsearch/src/elasticsearch/core.clj
```diff
@@ -220,9 +221,8 @@
                 (esi/flush client index-name)
                 (assoc op :type :ok
                        :value (->> (esd/search client index-name "number"
-                                               :search_type "query_then_fetch"
-                                               :scroll "1m"
-                                               :size 2)
+                                               :scroll "10s"
+                                               :size 20)
```
aphyr: Yes, that's much more sensible, thank you. :)

elasticsearch/src/elasticsearch/core.clj
```diff
@@ -243,16 +243,15 @@
...
       (try
         (esi/create client index-name
                     :mappings {mapping-type {:properties {}}})
-        (catch clojure.lang.ExceptionInfo e
-          ; Is this seriously how you're supposed to do idempotent
-          ; index creation? I've gotta be doing this wrong.
-          (let [err (http-error e)]
-            (when-not (re-find #"IndexAlreadyExistsException" err)
-              (throw (RuntimeException. err))))))
+        (catch Throwable t))
```
aphyr: Same here; I want to be conservative about which errors I'll ignore, haha. :)

elasticsearch/src/elasticsearch/core.clj
```diff
@@ -175,20 +178,18 @@
...
-          (let [err (http-error e)]
-            (when-not (re-find #"IndexAlreadyExistsException" err)
-              (throw (RuntimeException. err))))))
-
+                    :settings {"index" {"refresh_interval" "1s"}})
+        (catch Throwable t))
+      (esa/cluster-health client
+                          {:index [index-name] :level "indices"
+                           :wait_for_status "green"
+                           :wait_for_nodes "5"})
```
aphyr: Should "5" be a string instead of a number? Either way, we should use `(count (:nodes test))`.

elasticsearch/src/elasticsearch/core.clj
dakrone: Sure, I will remove this to only handle `IndexAlreadyExistsException` as you previously did. I do think using basic `clj-http` would be easier, as it'd allow you to use:

```clj
(try+
  ...
  (catch [:status 400]
    ;; ignore
```

elasticsearch/src/elasticsearch/core.clj
dakrone: It can be either a string or a number, I will change this to be `(count (:nodes test))` as you said.

dakrone:
Pushed another commit addressing your feedback, thanks for taking a look!

aphyr:
Excellent, thanks @dakrone :)

aphyr:
Do these tests run for you? Elasticsearch doesn't even start on my nodes any more; times out waiting for cluster recovery.

dakrone:
@aphyr I just double-checked this and the tests are still running for me

PR #52, opened 2015-04-22
Database tests fail to compile

jhalterman:
The various database tests (aerospike, elasticsearch, etc.) currently fail to compile, with something like:

```
Exception in thread "main" java.io.FileNotFoundException: Could not locate jepsen/db__init.class or jepsen/db.clj on classpath: , compiling:(elasticsearch/core.clj:1:1)
	at clojure.lang.Compiler.load(Compiler.java:7142)
```

It looks like the jepsen `0.0.3-SNAPSHOT` depedency is pulling down from clojars and is missing some required files. This PR just updates the jepsen dependency versions to `0.0.3`. Alternatively you may want to just change everything to 0.0.4-SNAPSHOT, but here's this for now.

### DIFF
--- a/aerospike/project.clj
+++ b/aerospike/project.clj
@@ -4,5 +4,5 @@
   :license {:name "Eclipse Public License"
             :url "http://www.eclipse.org/legal/epl-v10.html"}
   :dependencies [[org.clojure/clojure "1.6.0"]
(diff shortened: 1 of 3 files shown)


aphyr:
Bumped to 0.0.3

PR #56, opened 2015-05-20
Jepsen modularity fixes and extensions (SSH ports, NetController)

handlerbot:
If you have scrunchyface opinions about the netcontrol/NetControl and :netcontroller names, I'm totally open to changing them without too much complaint. I just saw that we had control/net.clj and net.clj and (reify net/Net) and net-the-module-shortname and :net-the-test-map-hash-key, all referencing slightly different things, and I got concerned about short names costing more than they benefitted.

### DIFF
--- a/jepsen/project.clj
+++ b/jepsen/project.clj
@@ -17,6 +17,7 @@
                   :exclusions [org.slf4j/slf4j-api
                                org.slf4j-log4j12
                                com.google.guava/guava]]]
+  :classifiers [["rabbitmq" :rabbitmq]]
   :profiles {:consul {:source-paths ["consul/src"]
                       :test-paths   ["consul/test"]
                       :dependencies [[cheshire "5.4.0"]
--- a/jepsen/src/jepsen/nemesis.clj
+++ b/jepsen/src/jepsen/nemesis.clj
@@ -4,8 +4,8 @@
             [jepsen.util        :as util]
             [jepsen.client      :as client]
             [jepsen.control     :as c]
-            [jepsen.control.net :as jcn]
-            [jepsen.netcontrol  :as netcontrol]))
+            [jepsen.control.net :as controlnet]
+            [jepsen.net         :as net]))
 
 (defn noop
   "Does nothing."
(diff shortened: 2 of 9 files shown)


aphyr:
I'd call this one 'net, and adopt 'control.net for anything that uses the old namespace. Ideally *nothing* should refer to control.net any more, because jepsen.net provides polymorphic funs for its features. :)

jepsen/project.clj
```diff
@@ -17,6 +17,7 @@
                   :exclusions [org.slf4j/slf4j-api
                                org.slf4j-log4j12
                                com.google.guava/guava]]]
+  :classifiers [["rabbitmq" :rabbitmq]]
```
aphyr: What's this for? Ideally I don't want Jepsen knowing RabbitMQ exists.

jepsen/src/jepsen/nemesis.clj
```diff
@@ -4,7 +4,8 @@
             [jepsen.util        :as util]
             [jepsen.client      :as client]
             [jepsen.control     :as c]
-            [jepsen.control.net :as net]))
+            [jepsen.control.net :as controlnet]
```
aphyr: I think control.net would be idiomatic here

jepsen/project.clj
handlerbot: This lets me do `(:require [jepsen.system.rabbitmq :as rmq])` in my core_test, so I can do `{:client  (rmq/queue-client)}` in my `test`, and reference the RabbitMQ client from your implementation rather than forking it into mine. You activate it by `:dependencies [jepsen "0.0.4-SNAPSHOT" :classifier "rabbitmq"]` in `project.clj`.

jepsen/src/jepsen/net.clj
```diff
@@ -0,0 +1,24 @@
+(ns jepsen.net
+  "Controls network manipulation."
+  (:use jepsen.control))
+
+(defprotocol Net
+  (drop-from! [net source] "Drop traffic originating from this source IP.")
```
aphyr: This implicitly assumes jepsen.control is telling us what nodes to act on. I'd prefer to make the source and dest nodes explicit; e.g. (drop! [net source dest])

jepsen/project.clj
handlerbot: (Basically, this lets me activate your `rabbitmq` profile from downstream inheritance; if I don't do that, I can't see `jepsen.system.rabbitmq` to `require` it.)

aphyr: Ah, yeah, that's still buried inside jepsen isn't it. I gotta pull that out into its own project. Classifier sounds good for now. :)

aphyr:
Looks basically good to go; thanks for taking a stab at this, @handlerbot! The only structural change I'd make is making source/dest nodes explicit in jepsen.net and having the iptables implementation use `control/on` to run the commands on that node.

jepsen/src/jepsen/nemesis.clj
handlerbot: Done.

jepsen/src/jepsen/net.clj
handlerbot: Done.

jepsen/src/jepsen/nemesis.clj
```diff
@@ -14,25 +14,20 @@
...
-  "Drops all packets from node."
-  [node]
-  (c/su (c/exec :iptables :-A :INPUT :-s (net/ip node) :-j :DROP)))
-
 (defn snub-nodes!
   "Drops all packets from the given nodes."
-  [nodes]
-  (dorun (map snub-node! nodes)))
+  [test dest sources]
+  (doseq [src sources] (net/drop! (:net test) test src dest)))
```
aphyr: OK I'm just gonna ask for oonnnnne more thing here, that I didn't think of before--we should still parallelize the cutoffs by writing this as (->> sources (pmap #(net/drop! (:net test) test % dest)) dorun)

handlerbot: Done.

aphyr:
Wooo, thanks! :)

PR #59, opened 2015-06-03
Make intermediate directories before writing gnuplot graph

jkni:
On first checkout and run, gnuplot would error when creating the latency graph because it won't create intermediate directories and the "report" directory had not yet been created using report/to.

The patch creates these intermediate directories; it has been tested with the elasticsearch tests.


aphyr:
Excellent!

PR #62, opened 2015-06-05
Update noop-test to use nemesis/noop

jkni:
Previously, noop-test used client/noop.

Client/noop assocs :type :ok on the argument op, which breaks the pairing in the timeline/html checker when used with noop-test.

This pull request updates noop-test to use nemesis/noop and makes nemesis/noop a def rather than a defn for consistency.


aphyr:
Thank you, @jkni :)

PR #66, opened 2015-06-26
Update Jepsen docker setup to support reuse of Jepsen containers.

jhalterman:
This change allows a jepsen container to be reused (after a stop/start) by re-running any needed sub-container setup at each login.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -16,7 +16,7 @@ To start run (note the required --privileged flag)
 Building the docker image
 =========================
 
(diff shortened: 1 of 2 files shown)


tjake:
This should also check if the containers are running already before trying to start them since this would error if you tried to connect to the main container twice (or created another shell)

jhalterman:
@tjake Thanks for the heads up. Updated the commit.

aphyr:
Thank you! :)

PR #72, opened 2015-08-16
Store the host keys unencrypted so that jsch can use them

stapelberg:


aphyr:
Yas kween! Thanks for helping streamline this process--I know Jepsen is a bear to set up. :)

PR #75, opened 2015-09-07
Add SmartOS support and example project

jkni:
This PR adds SmartOS support.

In particular, it adds a jepsen.os.smartos namespace that supports the os protocol and adds analogs for the functions in the jepsen.os.debian namespace. It also adds an ipfilter implementation of the net protocol.  Like the iptables implementation, the ipfilter implementation assumes it has complete control of firewalling on that machine.

In order to showcase the use of these namespaces and make testing possible, this PR also includes an adapted version of the mongodb Jepsen tests. These have been successfully run against SmartOS. The tests for the Jepsen project also pass, as expected.

This has been tested against version 15.2.0 of the base-64 SmartOS image.

### DIFF
--- a/jepsen/src/jepsen/net.clj
+++ b/jepsen/src/jepsen/net.clj
@@ -23,3 +23,12 @@
       (on-many (:nodes test) (su
                                (exec :iptables :-F :-w)
                                (exec :iptables :-X :-w))))))
+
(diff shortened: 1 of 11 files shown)


aphyr:
You magnificent bastard. <3

PR #78, opened 2015-09-20
Fix hardcoded node names

bogdando:
W/o this patch, nodes names cannot be changed
at the single place (net.clj)


aphyr:
Oh jeez, I think this may all be dead code--node names are set in the test now. Happy to merge though. :)

PR #80, opened 2015-10-04
Add a RobustIRC test.

stapelberg:
See http://robustirc.net/ for more information, specifically:

http://robustirc.net/docs/adminguide.html for setup/overview
http://robustirc.net/docs/robustsession.html for the protocol


stapelberg:
This is my first clojure code, so bear with me :). It’s largely copy&pasted from the existing tests.

I’m well aware this code is not yet perfect, and I’m looking for feedback on all aspects: the code can surely still be improved, and possibly the jepsen test itself might need some adjustments/improvements.

I looked at the existing checkers and think checker/set is the only applicable one for RobustIRC.

robustirc/src/jepsen/robustirc.clj
```diff
@@ -0,0 +1,230 @@
...
+                " -network_name=jepsen"
+                " -tls_cert_path=/tmp/cert.pem"
+                " -tls_ca_file=/tmp/cert.pem"
+                " -tls_key_path=/tmp/key.pem"
+                " -join=" (name (jepsen/primary test)) ":13001")]
+          (if (= node (jepsen/primary test))
+                (Thread/sleep 100)
+                (do
+                  (info node (str "running: " cmd))
+                  (c/ssh* {:cmd cmd})
```
aphyr: You don't have to do all the string escaping here yourself if you like--it's OK to use `(c/exec :start-stop-daemon :--start ...)`.

robustirc/src/jepsen/robustirc.clj
```diff
@@ -0,0 +1,230 @@
...
+           (gen/nemesis
+             (gen/seq (cycle [(gen/sleep 0)
+                              {:type :info, :f :start}
+                              (gen/sleep 10)
+                              {:type :info, :f :stop}])))
+           (gen/time-limit 30))
+      (gen/nemesis (gen/once {:type :info, :f :stop}))
+      (gen/sleep 5))))
+
+(defn createsession
```
aphyr: (I'm gonna merge, but just as a heads-up idiomatic Clojure style is to use hyphens, e.g. `create-session`. :))

robustirc/src/jepsen/robustirc.clj
```diff
@@ -0,0 +1,230 @@
...
+      (postmessage session "JOIN #jepsen")
+      (assoc this :node node
+                  :session session)))
+
+  (invoke! [this test op]
+    (try
+      (case (:f op)
+        :add (try
+                 (do
+                   (postmessage (-> session) (str "TOPIC #jepsen :" (:value op)))
```
aphyr: Heh, clever. :)

robustirc/src/jepsen/robustirc.clj
```diff
@@ -0,0 +1,230 @@
...
+       :content-type :json
+       :form-params {:Data ircmessage
+                     :ClientMessageId msgid}})))
+
+(defn getmessages
+  [session resume]
+  (let [c (async/chan)]
+    (do
+      (async/go
+	; TODO: handle org.apache.http.ConnectionClosedException which gets
```
aphyr: It may be easier to do this synchronously--Jepsen assumes all your client calls are synchronous anyway in order to provide scheduling invariants.

robustirc/src/jepsen/robustirc.clj
```diff
@@ -0,0 +1,230 @@
...
+      [checker :as checker]
+      [nemesis :as nemesis]
+      [generator :as gen]]
+    [jepsen.os.debian :as debian]
+    [cheshire.core            :as json])
+  (:import
+    [java.net URL]))
+
+(defn db [version]
+  "Rethinkdb (ignores version)."
```
aphyr: May want to rename this to "RobustIRC" since we're not testing RethinkDB here. :)

robustirc/src/jepsen/robustirc.clj
```diff
@@ -0,0 +1,230 @@
...
+      [control :as c :refer [|]]
+      [checker :as checker]
+      [nemesis :as nemesis]
+      [generator :as gen]]
+    [jepsen.os.debian :as debian]
+    [cheshire.core            :as json])
+  (:import
+    [java.net URL]))
+
+(defn db [version]
```
aphyr: If you won't use `version`, might as well not have a parameter for it, and you can drop it from the tests too. Or just use a GIT tag/SHA, and check it out in the repo prior to building.

aphyr:
Looks mostly good! The only things I need you to fix are the rethinkdb reference and the version--it's fine to either drop it or add a version checkout, your call. :)

robustirc/src/jepsen/robustirc.clj
stapelberg: Do you have a suggestion on how to do this synchronously? GetMessages is a never-ending HTTP request (see http://robustirc.net/docs/robustsession.html#getmessages) and I couldn’t find a way to read with a client-side timeout, but I might just have overlooked something. Or maybe there is a different way to structure the code that just doesn’t occur to me…?

An alternative I considered is to add a new parameter to RobustIRC itself which would make this call timeout on the server side, but decided against it due to the additional complexity in the server and also because it somewhat felt like cheating — I want the test to use the same code path that regular clients exercise.

robustirc/src/jepsen/robustirc.clj
stapelberg: Which way do you prefer? Should jepsen plugins use `version`?

stapelberg:
Thanks for the review! I’ll give you a chance to comment on my follow-ups before I make the changes you pointed out.

robustirc/src/jepsen/robustirc.clj
aphyr: You could use (jepsen.util/timeout 5000 (some-expr only-evaluated when-a-timeout-occurs) actual-work-here).

aphyr: But I don't mind the core.async version either, just figured it might be simpler :)

robustirc/src/jepsen/robustirc.clj
aphyr: Doesn't matter to me--I just parameterize it so I can test multiple versions easily, and it makes the test more reproducible because we target a specific version and not the latest. :)

robustirc/src/jepsen/robustirc.clj
stapelberg: The synchronous version is indeed much simpler, thanks for the jepsen.util/timeout hint!

robustirc/src/jepsen/robustirc.clj
stapelberg: Done.

robustirc/src/jepsen/robustirc.clj
stapelberg: Is there an advantage to using the `:` form?

robustirc/src/jepsen/robustirc.clj
stapelberg: Done.

robustirc/src/jepsen/robustirc.clj
stapelberg: Removed `version`.

stapelberg:
Made a bunch of changes, please take another look :).

robustirc/src/jepsen/robustirc.clj
aphyr: You're welcome to use strings, symbols, or keywords; they all work the same in this api. :foo is just a little shorter to type.

PR #82, opened 2015-10-20
Fix link to checker.clj

AkihiroSuda:
This should be applied together with #81.


aphyr:
what even is markdown ;)

AkihiroSuda:
@aphyr 
FYI, this tool is helpful for previewing github-flavored markdown :smile: 
https://github.com/joeyespo/grip

PR #84, opened 2015-10-20
Handle whitespace properly in string escaping.

paulsm:
This patch corrects behaviour of the string escaping used, for example, when passing slurped file content to an "echo" command for configuration or setup of a node. Previously, if the content didn't contain characters to trigger the wrapping of it in quotes, but contained newline characters, the escaping mechanism wouldn't quote a string.

e.g. if the content was:
```
line1 has spaces in it
line2doesnt
```
The old escape function would convert that to
```
"line1 has spaces in it
line2doesnt"
```
But if the content was
```
line1hasnospaces
line2doesnteither
````
Then the old escape function would not wrap it in quotes, while the changed one does.


aphyr:
Excellent call! Thank you. :)

PR #85, opened 2015-11-02
Prefix POSIX timestamp with @ when setting time via date command.

jhalterman:
Currently, the `clock-scrambler` fails to set local time on my (Debian 8.1) jepsen nodes. The failure is something like:

```
root@n1:~# date +%s -s 1446496762
date: invalid date '1446496762'
```

This PR adds an @ prefix to the timestamp to allow the command to succeed.


aphyr:
Thanks!

jhalterman:
Pretty sure that was the fasted merged PR in the history of github :)

PR #86, opened 2015-11-02
Clock scrambler should properly handle :start and :stop ops

jhalterman:
Clock scrambler should properly handle :start and :stop ops by assoc'ing the resulting time as the op :value.

Without this, Nemesis processes presently crash due to the following assertion:

https://github.com/aphyr/jepsen/blob/e6da77648b5d7b35b66555eecaf5331b7104b36e/jepsen/src/jepsen/core.clj#L143

### DIFF
--- a/jepsen/src/jepsen/nemesis.clj
+++ b/jepsen/src/jepsen/nemesis.clj
@@ -141,9 +141,10 @@
       this)
 
     (invoke! [this test op]
-      (c/on-many (:nodes test)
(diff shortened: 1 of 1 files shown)


aphyr:
Let's keep the assoc, but drop the start/stop logic--I planned on invoking this one over and over again and randomizing the clock each time.

jhalterman:
Sounds good - updated.

aphyr:
Woo, thanks!

PR #87, opened 2015-11-18
Update rabbitmq test to use 3.5.6

btobolaski:
Also fixes the teardown to allow multiple runs without manual intervention.


aphyr:
Thanks @btobolaski!

PR #89, opened 2015-12-31
Add test for LogCabin

siddontang:
A Jepsen test for [LogCabin](https://github.com/logcabin/logcabin) with test case like Jepsen Etcd and Consul.

Hi @aphyr 
Your post https://aphyr.com/posts/316-jepsen-etcd-and-consul says Etcd and Consul may have some linearizable problem, so I write a similar test for LogCabin and it can pass the test, but I am not sure whether my test is ok or I miss something with Jepsen (I am not familiar with clojure, only learn it for some days), I hope for your help to check the test :-) 

Thank you very much.

### DIFF
--- a/logcabin/.gitignore
+++ b/logcabin/.gitignore
@@ -1,0 +1,11 @@
+/target
+/classes
+/checkouts
+pom.xml
+pom.xml.asc
+*.jar
+*.class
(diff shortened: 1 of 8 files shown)


aphyr:
Cool, thank you!

PR #90, opened 2016-01-05
use go 1.5.2 to build Etcd.

siddontang:
Hi @aphyr 

Now the latest Etcd can't be built with default golang 1.3 with apt-get install, so I use the latest stable golang 1.5.2 instead.


aphyr:
Thanks sid!

PR #101, opened 2016-04-06
rabbitmq: remove noisy exceptions

bogdando:
Use with-ignore wrapper when we don't care
of exceptions (timeouts related mostly)


bogdando:
@aphyr I'm not sure why one may want to see those hundreds of timeout exceptions, so I suggested that patch...

PR #103, opened 2016-04-06
Fix iptables rules

bogdando:
On Nemesis start, drop related, established as well

Partially closes
https://github.com/aphyr/jepsen/issues/102

### DIFF
--- a/jepsen/src/jepsen/control/net.clj
+++ b/jepsen/src/jepsen/control/net.clj
@@ -57,7 +57,8 @@
 (defn cut-random-link
   "Cuts a random link to any of nodes."
   [nodes]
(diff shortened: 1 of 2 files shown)


bogdando:
sup :)

aphyr:
Related and established connections are *already* dropped. There's no reason to do this; it just allows through untracked and invalid packets, which is *not* what Jepsen wants to do.

bogdando:
The simple check may show that the blocking rule as-is will not affect some of connections and traffic will still flow after the nemesis start

bogdando:
I see, for the clean env you recommend in the setup guide, w/o accepting iptables rules, this works well. But if Jepsen framework is supposed to be used with arbitrary n1..n5 nodes, there *may* be ACCEPT rules for the cluster under test pre-created. Then, the rule will not match related traffic.

PR #105, opened 2016-04-07
Fix the checker throwing exception in analysis

bogdando:
It seems, the unexpected evaluator behaves in unexpected
way. Ignore exceptions raised to make things go smooth.

Closes: https://github.com/aphyr/jepsen/issues/104


aphyr:
Throwing away an exception and returning nil does not fix the exception: it just makes the error harder to debug. I'm not going to merge this, sorry.

PR #106, opened 2016-04-11
Update lxc.md to clarify setup instructions.

mvle:
Add steps to update keys for apt to verify packages in containers or else the setup phase would fail when running lein test.


aphyr:
Thanks!

PR #107, opened 2016-04-12
Do not cast core/set on multisets

bogdando:
Closes issue https://github.com/aphyr/jepsen/issues/104


bogdando:
@aphyr Works for me, please take a look

bogdando:
sup

aphyr:
Converting multisets to sets could result in tests passing which should have actually failed. This checker just isn't designed to deal with multisets; silently coercing them is probably not a good idea.

PR #108, opened 2016-04-18
Fix missed parameter

bogdando:
Use [ this test model history _ ] for Checker reify.
The 'ops' is missing, so the test thorws w/o the fix.
It seems not used though, so masked with "_".


bogdando:
sup

aphyr:
These changes will break these tests; Galera, for example, uses Jepsen 0.0.7, which doesn't use this function signature. Please stop.

bogdando:
Indeed, I switched to the 0.7 for my custom test, so the test stopped working without the fix. sorry for what. Just trying to to the best ;(

PR #111, opened 2016-04-23
jepsen.store: Ensure no concurrency issues for update-symlinks!

norton:
Update the latest symlinks after all files have been created.  This issue is not always reproducible but does happen occasionally when running a series of jepson tests.

### DIFF
--- a/jepsen/src/jepsen/store.clj
+++ b/jepsen/src/jepsen/store.clj
@@ -242,10 +242,10 @@
   [test]
   (->> [(future (write-results! test))
         (future (write-history! test))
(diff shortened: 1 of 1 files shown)


norton:
I have been testing with this commit for several days and the following error no longer happens.

```
Caused by: java.nio.file.NoSuchFileException: store/datomic-crash/latest
at sun.nio.fs.UnixException.translateToIOException (UnixException.java:86)
```

aphyr:
Very sensible, thank you. Sorry bout the delay, got really sick.

PR #112, opened 2016-04-26
Fix wget "tries" option

hptabster:
The option in wget should be "--tries" not "--retries"


aphyr:
Good catch; not sure how that made it through. Sorry!

PR #114, opened 2016-04-27
Use pkill instead of killall, as it is supported on all Linux variants.

hptabster:
Note - killall is not installed on an out of the box Ubuntu 14.04 host (not sure  about other Linux variants), it's in the psmisc package. It seems pkill is on the host. It might be better to support pkill than installing psmisc in jepsen/src/jepsen/os/debian.clj.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ that (along with SSH username, password, identity files, etc) is all definable
 in your test. The account you use on those boxes needs sudo access to set up
 DBs, control firewalls, etc.
 
-Be advised that tests may mess with clocks, add apt repos, run killall -9 on
+Be advised that tests may mess with clocks, add apt repos, run pkill -9 on
 processes, and generally break things, so you shouldn't, you know, point jepsen
 at your prod machines unless you like to live dangerously, or you wrote the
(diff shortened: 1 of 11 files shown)


aphyr:
I'm a little bit wary of running pkill blindly, because we might match unintended process names. (If we *do* want pattern matching, `control.util/grepkill` handles that). May I suggest adding something like `(c/exec :apt-get :install "psmisc")` to your OS `setup` function? It's a pretty cheap package, and I'm surprised it doesn't ship by default. Been an ubuntu user for a long while and... never found it missing, haha. Wonder if I got a different set of default packages somehow.

hptabster:
That's fair, would you consider a pull request which installs this package
as such:

       (install [:wget
                  :sysvinit-utils
                  :psmisc
                  :curl
                  :vim
                  :man-db
                  :faketime
                  :unzip
                  :iptables
                  :iputils-ping
                  :rsyslog
                  :logrotate])

aphyr:
Nooo problem, here, I'll do that right now. :)

aphyr:
And done. :)

PR #117, opened 2016-05-11
Include libsnmp30 for MongoDB Enterprise builds

hptabster:


aphyr:
Hi hptabster! It may be a better idea to just invoke (debian/install [:libsnmp30]) in the setup! fn for MongoDB, unless this dependency is likely to be used in multiple tests.

hptabster:
It seems this is a standard package, which should be included with the
other packages for that platform. Doesn't it make sense to do it there?
Perhaps other DBs will need this library. I do not see a mongodb specific
package install, except for mongodb binaries.

aphyr:
Yeah, I just try to keep the OS setup to stuff that *every* test is going to need, and install packages for specific DBs in the DB setup phase. Means you don't have to pull in a zillion packages to run a given test. :)

hptabster:
So, should I put in a pull request to do it out of the mongodb setup, or
will you add it?

hptabster:
PS - I tried to put the following into mongodb/src/jepsen/mongodb/core.clj
        (debian/install [:libsnmp30])
But it throws a runtime error:

ERROR 21:36:46 jepsen.mongodb.runner - Oh jeez, I'm sorry, Jepsen broke.
Here's why:
java.util.concurrent.ExecutionException: clojure.lang.ArityException: Wrong
number of args (2) passed to: debian/install

aphyr:
Hard to tell without context. Is it possible you inserted that line into a macro like -> or doto?
On May 15, 2016 8:37 PM, hptabster <notifications@github.com> wrote:PS - I tried to put the following into mongodb/src/jepsen/mongodb/core.clj
        (debian/install [:libsnmp30])
But it throws a runtime error:

ERROR 21:36:46 jepsen.mongodb.runner - Oh jeez, I'm sorry, Jepsen broke.
Here's why:
java.util.concurrent.ExecutionException: clojure.lang.ArityException: Wrong
number of args (2) passed to: debian/install

hptabster:
Yes:

(defn db
  "MongoDB for a particular HTTP URL"
  [url]
  (reify db/DB
    (setup! [_ test node]
      (doto node
        (debian/install [:libsnmp30])
        (install! url)
        (configure! test)
        (start!)
        (join! test)))

    (teardown! [_ test node]
      (wipe! node))))

aphyr:
Yup! Doto puts `node` as the first argument to `debian/install`, which is why it's complaining about getting 2 args instead of 1. Just move the install outside the doto. :)

hptabster:
Now done in another way.

PR #121, opened 2016-05-17
Fix unquoting for timed out ops

bogdando:


bogdando:
this fixes timeout corner cases to raise a correct expection

PR #124, opened 2016-05-24
Save mongod.log to /root before cleaning up directory

hptabster:


aphyr:
You know the logs are preserved in store/latest/n*/, right?

PR #130, opened 2016-06-17
Wait in killall until all processes actually die.

hptabster:
The mongod process may take sometime to die after receiving the kill signal. We have seen this issue where the subsequent mongod failed to start because the socket was already in use.

Using the -w switch for killall should ensure that it is not longer running. Note, the killall command could hang if the process never terminates!

If this is not an optimal solution, then I suggest the stop-daemon function be modified to wait for all signaled processes to be terminated.


aphyr:
Sounds good to me. We can always wrap in `util/timeout` if necessary. Thanks @hptabster!

PR #139, opened 2016-07-17
docker: add docker-compose.yml

AkihiroSuda:
The previous "Docker-in-Docker" image has not been working since last December:
the image depends on `kojiromike/dind` which contains Docker 1.5, but Docker 1.5
can no longer pull images from Docker Hub: https://blog.docker.com/2015/10/docker-hub-deprecation-1-5/

Basically we could solve the issue by just upgrading dind (https://hub.docker.com/_/docker/),
but I feel replacing dind with Docker Compose is better in its simplicity and extensibility.

PTAL @tjake @jkni @jhalterman


docker/control/bashrc
```diff
@@ -0,0 +1,14 @@
...
+Welcome to Jepsen on Docker
+===========================
+
+This container runs the Jepsen tests in sub-containers.
+
+You are currently in the base dir of the git repo for Jepsen.
+If you modify the core jepsen library make sure you "lein install" it so other tests can access.
+
+To run a test:
+   cd mongodb && lein run
```
AkihiroSuda: The reason why I mentioned `mongodb` here is mainly because its README.md is well-documented than others now

docker/up.sh
```diff
@@ -0,0 +1,46 @@
...
+    
+    INFO "Generating ./secret/control.env"
+    echo '# generated by jepsen/docker/up.sh, parsed by jepsen/docker/control/bashrc' > ./secret/control.env
+    echo '# NOTE: \\n is expressed as ↩' >> ./secret/control.env
+    echo SSH_PRIVATE_KEY="$(cat ./secret/id_rsa | perl -p -e 's/\n/↩/g')" >> ./secret/control.env
+    echo SSH_PUBLIC_KEY=$(cat ./secret/id_rsa.pub) >> ./secret/control.env
+    
+    INFO "Generating ./secret/node.env"
+    echo '# generated by jepsen/docker/up.sh, parsed by the "tutum/debian" docker image entrypoint script' > ./secret/node.env
+    echo ROOT_PASS=root >> ./secret/node.env
```
AkihiroSuda: I'm not sure why we need `ROOT_PASS` here.
Just keeping the original behavior.
ping @tjake

jhalterman:
Just tried this out and it works good.

I'm not really sure what the pros/cons are versus the dind approach. Any comment?

aphyr:
(I dunno anything about docker but I'm happy to merge whenever y'all have working consensus! Just let me know! :))

jkni:
Docker compose is probably the cleaner approach here - dind was the old
hack before Docker supported this properly through compose. I can't try
this until next week, but I'd agree that is the appropriate move in theory.

AkihiroSuda:
@jhalterman

Pros:

- You can easily use the latest Docker
- You can build the image in just one-line command. You no longer need to run `docker commit` manually. (please refer to the changelog of `docker/README.md`)
- init scripts are significantly simplified because Docker Compose automatically sets up DNS
- You can spread n1..n5 across a Docker Swarm cluster for balancing the load.

Cons:

- You need to additionally install Docker Compose. But I think this con is trivial because its installation is very easy.

jhalterman:
Sounds good. 👍

AkihiroSuda:
ping @tjake :smiley:

tjake:
:+1:

AkihiroSuda:
@tjake Thank you for the response.
Do you know why we need `ROOT_PASS` to be explicitly set? https://github.com/aphyr/jepsen/pull/139/commits/d97fc25d2e91cf4bfea0800b0534c8b6d341978f

Someone and I found that jsch fails with `com.jcraft.jsch.JSchException: Auth fail` if `ROOT_PASS` is not set, even though `ssh n1` .. `ssh n5` succeeds.

tjake:
I'm pretty sure that's how jepsen authenticates to the nodes

tjake:
Also see https://github.com/aphyr/jepsen#jsch-auth-errors

aphyr:
It'll use your ssh agent by default, if you've got keys loaded. Otherwise it falls back to user/pass.

AkihiroSuda:
Thank you all for the comments, and I think the PR is now mergeable

PR #149, opened 2016-08-11
MongoDB: Save /opt/mongodb/stdout.log to /root

hptabster:


aphyr:
Hey! Just as a headsup, these are also saved on the control node in store/latest/<node-name>/, so you can tar up the whole directory in one shot. :)

PR #156, opened 2016-10-16
Make debian/install also use :--force-yes when no version is specified

dralves:
When specific versions of packages are specified, debian/install
uses :--force-yes, but that is not the case when no version is specified.
This leads to installations failing for the latter case with:
"There are problems and -y was used without --force-yes"

This just adds the missing parameter.


aphyr:
Righto, thanks David. :)

PR #157, opened 2016-11-12
cockroach: better handle restarts, new macro with-restart

knz:
The "restart transaction" errors can be embedded in BatchUpdateExceptions as often as PGSQLExceptions, so use java.sql.SQLException to catch both.

### DIFF
--- a/cockroachdb/src/jepsen/cockroach/client.clj
+++ b/cockroachdb/src/jepsen/cockroach/client.clj
@@ -130,8 +130,9 @@
   converts them to :fails"
   [op & body]
   `(try ~@body
-       (catch org.postgresql.util.PSQLException e#
-         (if (re-find #"ERROR: restart transaction" (.getMessage e#))
+       (catch java.sql.SQLException e#
+         (if (re-find #"ERROR: restart transaction"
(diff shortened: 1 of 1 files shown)


knz:
Also this PR includes a new macro `with-restart` 
This does work (I tested it locally). I didn't push my changes to tests to use it because it hurts performance, see https://github.com/cockroachdb/cockroach/issues/5935, but it'd be nice to include it in the jepsen distribution for future reference.

tamird:
Shouldn't this be relying on the error code rather than doing string matching?

knz:
@tamird probably, but do you know how to reliably extract the pg error code from the SQLException?

bdarnell:
It looks like you use [exception.getSQLState](http://postgresql.nabble.com/SQLException-getErrorCode-td2175792.html). (The postgres driver doesn't implement getErrorCode because JDBC specifies that method returns an int, but postgres error codes are alphanumeric). I think you'll have to reach into the BatchUpdateException to get the underlying PSQLException. 

LGTM

aphyr:
Thanks @knz :)

PR #174, opened 2017-04-23
Fix crashing omission in tutorial

qwitwa:
The text says "Also note that where we modify the value--for instance, in
`:f :read`--we have to construct a special `independent/tuple` for the
key/value pair."
but the code sample omitted that, resulting in exceptions.


aphyr:
Thanks!

PR #178, opened 2017-06-12
cockroach: Reduce test flakiness

bdarnell:
The register tests were failing consistently because the concurrency parameter was not set correctly; the first commit fixes that.

The second commit modifies `util/timeout` to improve error reporting (when there was an `InterruptedException`, it was getting clobbered by a `NullPointerException` in `unwrap-exception`).

The third commit fixes a hang that is sometimes seen at shutdown in several tests (always associated with the `InterruptedException` noted above). I haven't quite figured out why this is so, but empirically it seems to work. 

cc @knz

### DIFF
--- a/jepsen/src/jepsen/util.clj
+++ b/jepsen/src/jepsen/util.clj
@@ -11,6 +11,7 @@
             [clojure.tools.logging :refer [debug info warn]]
             [knossos.history :as history])
   (:import (java.util.concurrent.locks LockSupport)
+           (java.util.concurrent ExecutionException)
            (java.io File
                     RandomAccessFile)))
 
@@ -262,27 +263,16 @@
   (info (with-out-str (pprint x)))
   x)
 
-(defmacro unwrap-exception
-  "Catches exceptions from body and re-throws their causes. Useful when you
-  don't want the wrapper from, say, a future's exception handler."
-  [& body]
-  `(try ~@body
-        (catch Throwable t#
-          (throw (.getCause t#)))))
-
 (defmacro timeout
   "Times out body after n millis, returning timeout-val."
   [millis timeout-val & body]
-  `(let [thread# (promise)
-         worker# (future
-                   (deliver thread# (Thread/currentThread))
-                   ~@body)
-         retval# (unwrap-exception
-                   (deref worker# ~millis ::timeout))]
+  `(let [worker# (future ~@body)
+         retval# (try
+                   (deref worker# ~millis ::timeout)
+                   (catch ExecutionException ee#
+                     (throw (.getCause ee#))))]
      (if (= retval# ::timeout)
-       (do ; Can never remember which does which
-           (.interrupt @thread#)
-           (future-cancel worker#)
+       (do (future-cancel worker#)
            ~timeout-val)
        retval#)))
 
--- a/cockroachdb/src/jepsen/cockroach/adya.clj
+++ b/cockroachdb/src/jepsen/cockroach/adya.clj
@@ -27,7 +27,7 @@
     (let [client (c/client node)]
       (locking table-created?
         (when (compare-and-set! table-created? false true)
-          (rc/with-conn [c client]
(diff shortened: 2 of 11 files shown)


bdarnell:
Added another commit to fix the `:concurrency` parameter to the `sequential` test, same as for `register`. Any idea why these tests are failing for us now when they were passing back in the fall? Is it that we're running with a different number of nodes and the default for `:concurrency` is derived from the number of nodes?

aphyr:
(made a comment on https://github.com/cockroachdb/cockroach/issues/15734)

aphyr:
As for why they're failing now... I feel like you should have had to pass --concurrency 10 (or whatever) since fall. I haven't patched anything in the cockroach tests specifically, but perhaps an... update to Jepsen or your CI scripts changed something?

knz:
Yes probably it's me making a change to the concurrency parameter without understanding that the value was restricted.

bdarnell:
Our CI scripts have never passed `--concurrency`, and they used to pass (we don't have logs going back far enough to see when they started failing). We could certainly pass `--concurrency` there, but it seems like a strange bit of knowledge-at-a-distance that the CI scripts need to know about a constraint arising from the hard-coded `10` in the `independent/concurrent-generator` call. (maybe the test should round the supplied value up instead of clobbering it with a hard-coded concurrency like i've done here).

aphyr:
That's OK too, but I don't think you want to use the default concurrency *anyway*. You can find bugs faster by raising the concurrency, at least until cockroach starts getting overwhelmed by the load or Jepsen starts hitting thread contention issues. Try `--concurrency 10n`?

bdarnell:
OK, we can just run with more concurrency. What configuration did you use when you were running the tests for the writeup?

I'll remove the concurrency commits from this PR and just leave the other two.

aphyr:
I usually ramp up and down concurrency depending on how much the cluster can take and to get a picture of the system under load vs quiet. Looking at the histories, I ran with anything from 10 to 300 in my tests, but take a look at the latency plots coming out of your CI cluster, because different tests have different costs. 30 might be a safe general-purpose option.

bdarnell:
Ping. The remaining two commits in this branch are still relevant, I think. The first one fixes error reporting, and the second fixes a hang at shutdown.

aphyr:
Yeah, I'm just waiting on your responses. I'm concerned that the chances to timeout could actually *cause* deadlocks...

bdarnell:
Where are the comments you're waiting on responses to? All I see in this thread is talk about the now-reverted `concurrency` changes.

aphyr:
The comments I made aren't visible?

screenshot from 2017-06-29 12-49-58

bdarnell:
They're still pending. You need to finish the review ("review changes" in upper right) to publish the comments.

aphyr:
I... what... fucking hell, Github. I don't understand why some comments go through immediately and others get into this weird staging-for-review state...

bdarnell:
"Single comments" go through immediately, "review comments" get held up and sent as a single batch.

aphyr:
Yeah. I just.... I made all of these as regular old line comments in the change view. No idea how they got stuck for weeks like that. So sorry.... :(

jepsen/src/jepsen/util.clj
```diff
@@ -262,27 +263,16 @@
...
-         retval# (unwrap-exception
-                   (deref worker# ~millis ::timeout))]
+  `(let [worker# (future ~@body)
+         retval# (try
+                   (deref worker# ~millis ::timeout)
+                   (catch ExecutionException ee#
+                     (throw (.getCause ee#))))]
      (if (= retval# ::timeout)
-       (do ; Can never remember which does which
-           (.interrupt @thread#)
```
bdarnell: `future-cancel` interrupts the worker thread for you. Added a test to verify this.

aphyr:
Right, good to go. :)

PR #181, opened 2017-06-27
Misc. fixups for cockroachdb tests

tafkacn:
- Upgrade to Jepsen 1.0.5
- Add parenthesis around linearizable checker in register test to get the actual checker instance rather than the function that returns the instance
- Updated cockroach/os/ubuntu.clj to use jepsen.net and fix-up heal call
- Added lein-localrepo plugin to jepsen library build so it is now possible to install the local snapshot build to local maven repository for testing.


aphyr:
Looks good! 0.1.6-SNAPSHOT has the improved linearizability checker, BTW. :)

PR #193, opened 2017-08-11
Hazelcast: trivial improvements

vjuranek:


hazelcast/src/jepsen/hazelcast.clj
```diff
@@ -35,7 +35,9 @@
 (def local-server-jar
   "Relative to server fat jar"
   (str local-server-dir
-       "/target/jepsen.hazelcast-server-0.1.0-SNAPSHOT-standalone.jar"))
+       "/target/jepsen.hazelcast-server-"
+       (System/getProperty "projectname.version")
```
aphyr: Since this is not the same project, their versions aren't necessarily going to be the same. Might be more robust to look at the target directory contents instead.

hazelcast/src/jepsen/hazelcast.clj
```diff
@@ -347,7 +349,8 @@
                        :checker  (checker/unique-ids)}
     :id-gen-ids       {:client    (id-gen-id-client nil nil)
                        :generator {:type :invoke, :f :generate}
-                       :checker   (checker/unique-ids)}))
+                       :checker   (checker/unique-ids)}
+    (throw (IllegalArgumentException. "Empty/illegal '--workload' argument"))))
```
aphyr: Gonna merge this, but I suggest baking it into opt-spec instead.

aphyr:
Thanks @vjuranek!

vjuranek:
Thanks! You are right, I'll try to address your comments later this week. Feel free to keep the PR open until I fix the issues next time :-)

aphyr:
It looks like your changes to server jarfiles broke compilation: I get `java.util.concurrent.ExecutionException: java.io.FileNotFoundException: /home/aphyr/jepsen/hazelcast/server/target/jepsen.hazelcast-server--standalone.jar (No such file or directory)`; I think because the system property you're reading doesn't exist. Would you mind either reverting or fixing this?

aphyr:
Whoops, looks like your fix was actually a part of #196 and I got so caught up on the input validation stuff I missed the compilation profile changes. I'll port those over.

PR #194, opened 2017-08-15
Add TiDB jepsen test

siddontang:
Hi @aphyr 

We have been building a distributed transactional database [TiDB](https://github.com/pingcap/tidb) and use Jepsen to verify its linearizability.


aphyr:
Nice work!

PR #196, opened 2017-08-19
Fixes for PR #193

vjuranek:
Improving changes done in https://github.com/jepsen-io/jepsen/pull/193

### DIFF
--- a/hazelcast/src/jepsen/hazelcast.clj
+++ b/hazelcast/src/jepsen/hazelcast.clj
@@ -35,9 +35,7 @@
 (def local-server-jar
   "Relative to server fat jar"
   (str local-server-dir
-       "/target/jepsen.hazelcast-server-"
-       (System/getProperty "projectname.version")
-       "-standalone.jar"))
+       "/target/hazelcast-server.jar"))
 
 (def dir
   "Remote path for hazelcast stuff"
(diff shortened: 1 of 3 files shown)


hazelcast/src/jepsen/hazelcast.clj
```diff
@@ -386,15 +388,24 @@
...
+  []
+  (str/join ", "(map name (keys (tests nil)))))
+
+
 (def opt-spec
   "Additional command line options"
-  [[nil "--workload WORKLOAD" "Test workload to run, e.g. atomic-long-ids."
-    :parse-fn keyword]])
+  [[nil "--workload WORKLOAD" (str "Test workload to run, possible options are " (test-names) "Required.")
+    :parse-fn keyword
```
aphyr: May I suggest `:parse-fn (comp (tests) keyword)`, and `:validate [identity (jepsen.cli/one-of (tests))]`? Then you don't need any of the changes to jepsen.cli, and workload lookup is handled by the option parser for you.

aphyr: You can also rename `tests` to `workloads` since it sort of describes workloads, not full tests. :)

vjuranek:
Thanks for the review!
Not sure if I get you right (I'm Jepsen as well as Clojure beginner), proposed changes are implemented [here](https://github.com/vjuranek/jepsen/commit/277ce5ef664a6101707e3c43317db630c76b72ed#diff-1160de51de9475e2447c69d89a13ced9R400), but it doesn't solve the main issue I wanted to fix: when you go to Hazelcast and run `lein run` it gives you a hint that one should run either server or tests, but running `lein run test` results into not very clear error
```
[main] ERROR jepsen.cli - Oh jeez, I'm sorry, Jepsen broke. Here's why:
java.lang.IllegalArgumentException: No matching clause:
```
or NPE in when adding only validation function HC CLI options. I wanted to provide similar comfort for the user to get help immediately, something like
```
$ lein run test                                                                                                                                                                                        :fix_pr
Missing one or more required argument(s). Required arguments are "workload". Type cmd --help for more details.
```

hazelcast/src/jepsen/hazelcast.clj
```diff
@@ -386,15 +388,24 @@
...
+  []
+  (str/join ", " (map name (keys (workloads nil)))))
+
+
 (def opt-spec
   "Additional command line options"
-  [[nil "--workload WORKLOAD" "Test workload to run, e.g. atomic-long-ids."
-    :parse-fn keyword]])
+  [[nil "--workload WORKLOAD" (str "Test workload to run, possible options are " (workload-names) "Required.")
+    :parse-fn keyword
```
aphyr: Like I mentioned in the previous review, you want to use `(comp (workloads) keyword)` as the parse fn. That way the option parser will look up the workload for you.

aphyr: Oh, wait, I think I'm looking at the wrong version of the code here. Ummmmm... hmm. This might be because (workloads) is a function of test, and we're passing in nil. I'll go poke at the code a bit.

aphyr:
Right, fixed these in 9ef7e1f738856be5d247395d0f4cf8b891ddebf7. Forgot that tools.cli uses `:missing`, not `:validate`, for checking mandatory args.

PR #203, opened 2017-10-16
Update nemesis.md

skillness:


aphyr:
Nemeses is the plural form of nemesis.

PR #210, opened 2017-10-26
Update CrateDB tests 

andreidan:
This PR upgrades the CrateDB tests to allow for installing Crate using a tarball and, as we don't support the client anymore, use jdbc to interact with CrateDB (we used https://github.com/clojure/java.jdbc/ )
We also added a few scripts for convenience, in order to run all tests or a specific one, and a short guide on how to trigger the tests and what to expect in the results section.

A few details on the changes: 
- the crate.yml structure has changed quite considerably since crate 0.55.2, which was the version tested previously, so the diff for the file is fairly radical, but the configuration options are essentially the same (except for the recovery ones)
- there was a bit of confusion as before as to what `crate.clj` is. This lived under src/jepsen and it  defined the version diversion test and also some utilities used by the `dirty-read` and `lost-updates` tests.
This was split into `version-divergence.clj` (for the specific test) and `core.clj` (for the crate setup/install and es utilities). Everything now lives under `src/jepsen/crate`
- `runner.clj` handles the parsing of arguments, test command and the main entry point.


aphyr:
Happy to merge this, but you should be aware that writing JDBC tests this way may result in some subtle concurrency bugs depending on test semantics, and you might see false positives or negatives. You probably want to follow something like the cockroach tests, using a jepsen.reconnect wrapper for stateful connection management.

andreidan:
Thanks Kyle, will have a look and use jepsen.rc

PR #214, opened 2017-10-30
Extend jepsen.client and jepsen.core with open! and close!

mkcp:
A notable addition since last review is that I've made sure the latest client used in the worker is used for closing. Overall the complexity of workers has gotten worse, and I've got the feeling there's more improvements and simplifications to be made there. But for now this gets us the open! and close! behavior, along with n (where n is :concurrency) setup and teardown calls before and after our invocation phase respectively.


jepsen/src/jepsen/client.clj
```diff
@@ -1,20 +1,69 @@
...
+    (setup!    [this test])
     (teardown! [this test])
-    (invoke!   [this test op] (assoc op :type :ok))))
+    (invoke!   [this test op] (assoc op :type :ok))
+    (open!     [this test node] this)
+    (close!    [this test node])))
+
+(defn reopen!
+  "Takes the worker's current client, closes its connection and
+  opens a new connection with the original."
```
aphyr: The docstring says this opens a new connection with the original, but it looks like it opens the current client again. Maybe... drop this function and use close!/open! directly in jepsen.core?

jepsen/src/jepsen/client.clj
```diff
@@ -1,20 +1,69 @@
...
+(defn open-compat!
+  "Attempts to call `open!` on the given client. If `open!` does not
+  exist, we assume a legacy implementation of `setup!`."
+  [client test node]
+  (try
+    (let [client (open! client test node)
+          _      (setup! client test)]
+      client)
+    (catch java.lang.AbstractMethodError e
+      (warn "DEPRECATED: `jepsen.client/open!` not implemented. Falling back to deprecated semantics of `jepsen.client/setup!`.")
```
aphyr: Let's also suggest a path forward to users. Maybe something like "You should separate your client's setup! function into open! and setup!. See the jepsen.client documentation for details."?

jepsen/src/jepsen/client.clj
```diff
@@ -1,20 +1,69 @@
...
+       (some #{'close_BANG_})))
+
+(defn close-compat!
+  "Inspects the client for `close!` method and calls `teardown!` then `close!`.
+  If `close!` is not implemented, we assume a legacy implementation of `teardown!`."
+  [client test node]
+  (if (closable? client)
+    (do (teardown! client test)
+        (close! client test node))
+    (do (warn "DEPRECATED: `jepsen.client/close!` not implemented. Falling back to deprecated semantics of `jepsen.client/teardown!`.")
```
aphyr: Same here; let's tell users they should implement a close! function, and refer them to client docs.

jepsen/src/jepsen/core.clj
```diff
@@ -138,72 +138,122 @@
...
+                                    (.getMessage t))))]
+        ; Add synthetic completion to history
+        (conj-op! test op))
+
+      (warn t "Invocation on process" process "indeterminate")
+
+      (if (client/closable? client)
+        ; New process, new client
+        [(+ process (:concurrency test))
+         (client/reopen! (:client test) test node)]
```
aphyr: This reads so well. :)

jepsen/src/jepsen/core.clj
```diff
@@ -138,72 +138,122 @@
...
+        ; Differentiate our base client for this process
+        (let [client (client/open-compat! (:client test) test node)
+              _      (assert client (str "Expected a client, but got " (pr-str client)
+                                         " instead."))
+
+              ; Wait to begin ops until all clients are set up
+              _      (.await setup-barrier)
+
+              ; Consume operations from the generator, returns most recent client when done
+              client (loop [[process client] [process client]]
```
aphyr: We're close here, but this section can still throw in a few ways--notably, the assert will throw an AssertionError and bail out immediately without closing the client. What about... wrapping the whole thing in a `try`, and using (finally (.await setup-barrier ...)) to ensure cleanup happens regardless of exceptions?

aphyr: (I recognize this is A Lot Of Nesting (TM)--feel free to break up into more fns as you see fit!)

jepsen/src/jepsen/core.clj
```diff
@@ -138,72 +138,122 @@
...
+                       ; Obtain an operation to execute
+                       (if-let [op (try (generator/op gen test process)
+                                        (catch Exception e e))]
```
aphyr: I think it might be more helpful for users to see a specific error message here, and re-throw (so we don't try to use the exception as an op). Something like `(catch Exception e (warn e "Generator X threw") (throw e))`, which will print out a stacktrace.

jepsen/src/jepsen/core.clj
```diff
@@ -215,7 +265,8 @@
       (with-thread-name "jepsen nemesis"
         (info "Nemesis starting")
         (loop []
-          (when-let [op (generator/op gen test :nemesis)]
+          (when-let [op (try (generator/op gen test :nemesis)
+                             (catch Exception e e))]
```
aphyr: Same here. Perhaps it's worth moving exception logging & validation (e.g. ops are maps with these keys) into their own dedicated functions in jepsen.generator?

jepsen/src/jepsen/core.clj
```diff
@@ -257,7 +308,7 @@
   nemesis completion, and tears down nemesis."
   [test & body]
   ; Initialize nemesis
-  `(let [nemesis# (client/setup! (:nemesis ~test) ~test nil)]
+  `(let [nemesis# (client/open-compat! (:nemesis ~test) ~test nil)]
```
aphyr: Shoot, I didn't even think about this. Nemeses aren't bound to any particular node, so `open!` doesn't make... logical sense for them. In fact, we call setup! with node *nil*. Do you think we should make nemeses their own protocol, since they don't really follow the same lifecycle as clients anyway?

jepsen/src/jepsen/core.clj
```diff
@@ -282,27 +333,23 @@
...
+      (let [setup-barrier (java.util.concurrent.CyclicBarrier. (:concurrency test))
+            client-nodes  (if (empty? (:nodes test))
+                            ; If you've specified an empty node set, we'll still
+                            ; give you `concurrency` clients, with nil.
+                            (repeat (:concurrency test) nil)
+                            (->> test
+                                 :nodes
+                                 cycle
+                                 (take (:concurrency test))))
+            workers       (mapv (partial worker test setup-barrier)
```
aphyr: Yeah, this is super well done. Nice job.

jepsen/src/jepsen/core.clj
mkcp: I agree, thanks! Extracted both cases out into `jepsen.generator/op-and-validate` and re-threw  with additional context.

jepsen/src/jepsen/core.clj
mkcp: Giving nemeses their own protocol makes a lot of sense to me. They get their own worker setup before all of these codepaths, so sharing the protocol here makes them intertwine in odd ways that were a constant edge case while implementing.

I can scope this out and ship it before we land 1.7?

jepsen/src/jepsen/client.clj
mkcp: Good idea :D this is now implemented for both deprecated messages

jepsen/src/jepsen/client.clj
mkcp: Yeah, this particular abstraction didn't win us much. It's been excised

jepsen/src/jepsen/core.clj
mkcp: Thank you for catching this! I believe this has been addressed in the updated mechanics.Took a bit of fiddling, but I got the clj compiler to be happy with the arrangement of try/catch/finally. I feel like a `with-client` macro could address some of the repetition here, but I'm not sure how much the indirection would win us in the single case. Additional feedback on how to improve this section would be appreciated.

jepsen/src/jepsen/core.clj
aphyr: Yeah, I think you're right. Let's do a Nemesis protocol, but fall back to calling client/setup! and client/teardown! when nemesis funs aren't available on a nemesis, for compatibility. Same story as the new Client functions.

jepsen/src/jepsen/core.clj
```diff
@@ -138,72 +138,126 @@
...
+                             (recur (invoke-and-complete node process client test op)))
+
+                           ; Out of ops, return most recent client
+                           client))
+
+                       ; Exceptions here aren't expected, clean up state then re-throw
+                       (catch Exception e
+                         (warn "Worker for" process "threw unexpectedly. Tearing down")
+                         (.await setup-barrier)
+                         (client/close-compat! client test)
```
aphyr: So the client that we tear down here might be a *different* client from the one inside the loop/recur. What about... maybe having a third loop/recur argument, which is nil normally, and an exception if we throw? Then we could do

```clj
(let [[process client exception]
  (loop [[process client exception] [process client nil]]
     (try ... (catch Exception e [process client e]))))
```

mkcp: You're right that the wrong client might get closed out here... I'd need a result from the loop even though it threw dang.

When attempting to pass the exception up for later, I'm having a difficult time getting around 'cannot recur across try'. I'm having a difficult time thinking around this, any ideas?

mkcp: Ok, got something with (recur (try ...)) and returning a bool to signal our exit condition

aphyr:
Bingo, I think this will do it! :)

PR #216, opened 2017-11-01
Nemesis protocol

mkcp:
The proof of concept was tested with `nemesis/partitioner`, which has been migrated in this branch. I also composed `partitioner` with `hammer-time` to verify that nemeses can be composed across protocols. After some research, I don’t believe any nemeses use the node arg to setup, they all get their node info from test — therefore node can be safely removed. Aside from removing the extra arg, migrating to the new protocol should be a drop-in replacement.

I opted not to warn when invoke is called to avoid spamming up the invocation history. Warning in the `setup!` and `teardown!` phases felt sufficient.

Once both this and the updates to jepsen.client land, I’d like to circle back to `jepsen.core/nemesis-worker` and update `generator/op` with `generator/op-and-validate`

### DIFF
--- a/jepsen/src/jepsen/core.clj
+++ b/jepsen/src/jepsen/core.clj
@@ -28,6 +28,7 @@
             [jepsen.generator :as generator]
             [jepsen.checker :as checker]
             [jepsen.client :as client]
+            [jepsen.nemesis :as nemesis]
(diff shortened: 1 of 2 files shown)


aphyr:
Looks great Kit!

PR #220, opened 2017-11-11
Timeline ops can be linked to and now link to themselves

mkcp:
Adds monotonically increasing ids to all ops in the history, and adds the current independent subhistory's history-key to each checker's opts. Also adds breadcrumbs to the timeline

New behavior works like:

### DIFF
--- a/jepsen/src/jepsen/checker/timeline.clj
+++ b/jepsen/src/jepsen/checker/timeline.clj
@@ -71,19 +71,19 @@
         op (or stop start)
         s {:width  col-width
            :left   (* gutter-width (get process-index p))
-           :top    (* height (:t-index start))}]
+           :top    (* height (:sub-index start))}]
     [:a {:href (str "#i" (:index op))}
      [:div {:class (str "op " (name (:type op)))
             :id (str "i" (:index op))
             :style (style (cond (= :info (:type stop))
                                 (assoc s :height (* height
                                                     (- (inc (count history))
-                                                       (:t-index start))))
+                                                       (:sub-index start))))
 
                                 stop
                                 (assoc s :height (* height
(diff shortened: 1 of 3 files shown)


jepsen/src/jepsen/checker/timeline.clj
```diff
@@ -90,6 +118,14 @@
        (reduce (fn [m p] (assoc m p (count m)))
                {})))
 
+(defn t-index
```
aphyr: We can probably just call this :index, since it's a simple integer sequence. I thought, reading this PR, that t-index meant *time* index somehow. :)

mkcp: I meant it as :timeline-index, though really it marks the subhistory... what do you think about :sub-index instead?  I use :index [here](https://github.com/jepsen-io/jepsen/pull/220/files#diff-e4e16e33aa65970d1b55ec53a96ca2a2R481) for position in the full history, and it's where I get the linkable element ids.

mkcp: We could also skip the extra state and derive subhistory index from the index of the first non-nemesis op in the subhistory, but that's maybe overcomplicating it

aphyr: Oh, I see! I was thinking you could use :index for everything, since it'll be a unique ID in any subhistory.

mkcp: Thinking abt this more, I want to go with just :sub-index. Deriving subhistory index at each op and opens up a bunch of interactions with nemeses, and because we have the whole subhistory right at the beginning and it's simple to precompute it and pass it forward.

PR #223, opened 2017-11-17
Speed up multiple visits of jepsen.web index by ~75%

mkcp:
Implemented memoized versions of `jepsen.store/load-results` and `jepsen.store/load`. Together they provide about a 75% decrease in repeated page loads. The bulk of the savings is in `load-results` but `load` provides a noticeable bump as well. (based on a bunch of eyeballing page refreshes, haven't run a statistically valid benchmark) Because caches are each per test-name/test-time, additional tests/results get appended to the cache without needing the refresh the whole thing. Load once, run n analyses, profit \o/

Some #s with 830 tests:
Initial pre-cached run
memo pre-cache
Additional runs:
screen shot 2017-11-16 at 4 25 40 pm

### DIFF
--- a/jepsen/src/jepsen/store.clj
+++ b/jepsen/src/jepsen/store.clj
@@ -170,6 +170,8 @@
     (let [in (fress/create-reader file :handlers read-handlers)]
       (fress/read-object in))))
 
+(def memoized-load (memoize load))
+
 (defn load-results
   "Loads only a results.edn by name and time."
   [test-name test-time]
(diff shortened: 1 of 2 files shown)


aphyr:
Shoot, forgot to review this, sorry! So this is a good idea, but some tests might be *super* big (GB?) and we can only load one or two at a time; this means that every test we load will consume Jepsen memory forever. Let's drop the cache for `load` on this PR, and that'll be a good start; maybe later we can cache fast-tests directly, since we don't need to retain the full results of each test, just the validity?

mkcp:
Continuing this PR in https://github.com/jepsen-io/jepsen/pull/230. This PR got orphaned cause its fork got vaped and remade. Changes requested pushed to the above PR.

PR #224, opened 2017-11-17
cli: Make tarball-opt backwards-compatible

bdarnell:
The move to jepsen 0.1.6 broke our scripts that used the `--tarball` option; this looks like what was intended (and the other users of `tarball-opt` are also on jepsen 0.1.5).


aphyr:
ooof, sorry bout that :-O

PR #227, opened 2017-11-17
charybdefs: New subproject for disk fault injection

bdarnell:
Currently builds and installs everything and provides some basic
controls for injecting faults. Future work includes building an
actual nemesis with this and providing more fine-grained controls.

I started doing this in a separate project because it seems likely to grow additional dependencies that you may not want in other packages (such as thrift support, which will be needed for the aforementioned fine-grained controls), but I can move it under the main `jepsen` directory if you'd rather. 

All feedback about package structure and general clojure-ness welcome.

### DIFF
--- a/charybdefs/project.clj
+++ b/charybdefs/project.clj
@@ -1,0 +1,8 @@
+(defproject jepsen-charybdefs "0.1.0-SNAPSHOT"
+  :description "charybdefs wrapper for use in jepsen"
+  :url "https://github.com/jepsen.io/jepsen"
+  :license {:name "Eclipse Public License"
+            :url "http://www.eclipse.org/legal/epl-v10.html"}
+  :dependencies [[org.clojure/clojure "1.8.0"]
+                 [jepsen "0.1.6"]
+                 [yogthos/config "0.8"]])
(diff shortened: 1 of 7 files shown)


charybdefs/project.clj
```diff
@@ -0,0 +1,8 @@
+(defproject jepsen-charybdefs "0.1.0-SNAPSHOT"
```
aphyr: Let's do "jepsen.charybdefs" instead

charybdefs/project.clj
```diff
@@ -0,0 +1,8 @@
...
+  :description "charybdefs wrapper for use in jepsen"
+  :url "https://github.com/jepsen.io/jepsen"
+  :license {:name "Eclipse Public License"
+            :url "http://www.eclipse.org/legal/epl-v10.html"}
+  :dependencies [[org.clojure/clojure "1.8.0"]
```
aphyr: I think you can leave off the Clojure dependency here (move it to a :dev profile), since this is gonna be a library, not a standalone app.

aphyr:
oh my god you've done it

I spent so long trying to get this running, THANK YOU for getting things built out. Delighted to fold this in. Let me know how it works out in your tests--when you've got a release you're happy with I'll sign and deploy it to Clojars for you.

PR #229, opened 2017-11-20
Control improvements

mkcp:
Various items off of plan.md

- Add host to with-trace output
- jepsen.control/upload can now take local java.io.Files and returns the remote-path when finished
- Log additional debugging context when jepsen.control/session fails, incl. hostname, port, username, and password per node.

New behavior in control/session 
When an SSH connection fails, we print context once per node and finally throw the first exception returned from `with-resources`'s `real-pmap`

Here we provide an invalid username, password, and hostname (3 context prints are above the fold here)
screen shot 2017-11-20 at 2 48 15 pm

### DIFF
--- a/jepsen/src/jepsen/control.clj
+++ b/jepsen/src/jepsen/control.clj
@@ -203,8 +203,8 @@
   (with-retry [tries *retries*]
     (rc/with-conn [s *session*]
       (let [local-paths (if (sequential? local-paths)
-                          (map ensure-path local-paths)
-                          (ensure-path local-paths))]
+                          (map file->path local-paths)
+                          (file->path local-paths))]
         (apply ssh/scp-to s local-paths remote-path remaining)
         remote-path))
     (catch com.jcraft.jsch.JSchException e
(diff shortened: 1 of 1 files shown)


jepsen/src/jepsen/control.clj
```diff
@@ -270,23 +284,29 @@
...
-  (rc/open!
-    (rc/wrapper {:open    (if *dummy*
-                            (fn [] [:dummy host])
-                            (fn [] (clj-ssh-session host)))
-                 :name    [:control host]
-                 :close   (if *dummy*
-                            identity
-                            ssh/disconnect)
-                 :log?    true})))
+  (binding [*host* host]
```
aphyr: Okay, so... this part I'm a little bit less sure about. I need to take a closer look at this code because I'm not super familiar with it any more, but rc/wrapper is actually *reopenable*, so I'm not sure whether it's a good idea to catch exceptions only on the *first* call to rc/open!. Could subsequent reopenings also throw? Perhaps... it would be a good idea to put the try/catch *inside* the open function, right around `(clj-ssh-session host)`, and also to only bind *host* for the call to (debug-data) there, so that we can remember the host for subsequent open attempts.

aphyr:
Wooo! Let's do it! :)

PR #235, opened 2017-12-21
charybdefs: Introduce a nemesis, use it in cockroachdb

bdarnell:
Aside from a really weird problem with `rm -rf`, this seems to be working. It's mostly equivalent to the `kill` nemesis for cockroach since IO errors (mostly) cause the process to panic. Don't do a release with this yet - I'm going to rename `charybdefs/nemesis` to introduce a handful of different nemeses (like one that lets reads work but returns ENOSPC for any write).

### DIFF
--- a/charybdefs/src/jepsen/charybdefs.clj
+++ b/charybdefs/src/jepsen/charybdefs.clj
@@ -2,6 +2,7 @@
   (:require [clojure.tools.logging :refer [info]]
             [jepsen.control :as c]
             [jepsen.control.util :as cu]
+            [jepsen.client :as client]
             [jepsen.os.debian :as debian]))
 
 (defn install-thrift!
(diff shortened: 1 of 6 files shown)


bdarnell:
> this seems to be working

I spoke too soon. It works when I run one test at a time, but one test run can still interfere with the next. I'm still poking at getting the cleanup process right before/after a run.

bdarnell:
OK, got that sorted out (shell expansion was tripping me up). Now we've successfully run our whole test suite with this nemesis.

aphyr:
Woooo! OK, one small change and I think we'll be good to go--can you move this to jepsen.nemesis.charybdefs? Trying to move all the various nemesis-related stuff under the nemesis ns. :)

bdarnell:
So move `charybdefs/src/jepsen/charybdefs.clj` to `charybdefs/src/jepsen/nemesis/charybdefs.clj`, or move it under `jepsen/src/jepsen/nemesis/` with the other nemeses? I'm not sure what's normal or allowed for clojure packages. Putting it under the main jepsen package would be a problem for us because it's sometimes hard for us to adjust to changes in that package (the cockroachdb tests are currently broken in cryptic ways if I use a checkout to link against the master version of jepsen)

aphyr:
Under the main Jepsen tree, please; I think it'll be useful for every test. :) If you're having trouble upgrading, let me know what errors you're seeing and I'll help you get them sorted out!

PR #237, opened 2018-03-02
Add support for running Jepsen without using SSH

visemet:
This enables tests which have the means to run without any OS-level virtualization to avoid attempting to establish SSH connections on start-up.

----

@aphyr, my changes from jepsen-io/mongodb#9 depend on these changes to the jepsen-io/jepsen repository.

### DIFF
--- a/jepsen/src/jepsen/control.clj
+++ b/jepsen/src/jepsen/control.clj
@@ -10,6 +10,7 @@
             [jepsen.reconnect :as rc]
             [clojure.string :as str]
             [clojure.pprint :refer [pprint]]
+            [clojure.java.shell :as shell]
             [clojure.tools.logging :refer [warn info debug error]]))
 
 ; STATE STATE STATE STATE
@@ -25,6 +26,9 @@
 (def ^:dynamic *private-key-path*         "SSH identity file"     nil)
 (def ^:dynamic *strict-host-key-checking* "Verify SSH host keys"  :yes)
 (def ^:dynamic *retries*  "How many times to retry conns"   5)
+(def ^:dynamic *run-locally*
+  "Whether to run commands on the local host rather than the remote host"
+  false)
 
 (defn debug-data
   "Construct a map of SSH data for debugging purposes."
@@ -100,17 +104,32 @@
   "Wraps command in a sudo subshell."
   [cmd]
   (if *sudo*
-    (merge cmd {:cmd (str "sudo -S -u " *sudo* " bash -c " (escape (:cmd cmd)))
+    (merge cmd {:cmd `("sudo" "-S" "-u" ~*sudo* "--" ~@(:cmd cmd))
                 :in  (if *password*
                        (str *password* "\n" (:in cmd))
                        (:in cmd))})
     cmd))
 
+(defn wrap-bash
+  "Wraps command in a bash shell."
+  [cmd]
+  (assoc cmd :cmd `("bash" "-c" ~(str/join " " (:cmd cmd)))))
+
+(defn wrap-bash-cd
+  "Wraps command in a bash shell prepended with a cd command."
+  [cmd]
+  (if (:dir cmd)
+    (-> cmd
+        (assoc :cmd `("bash" "-c" ~(escape (str "cd " (escape (:dir cmd)) "; "
+                                                (str/join " " (:cmd cmd))))))
+        (dissoc :dir))
+    (wrap-bash cmd)))
+
 (defn wrap-cd
   "Wraps command by changing to the current bound directory first."
   [cmd]
   (if *dir*
-    (assoc cmd :cmd (str "cd " (escape *dir*) "; " (:cmd cmd)))
+    (assoc cmd :dir *dir*)
     cmd))
 
 (defn wrap-trace
@@ -141,7 +160,7 @@
 (defn ssh*
   "Evaluates an SSH action against the current host. Retries packet corrupt
   errors."
-  [action]
+  [{cmd :cmd, :as action}]
   (with-retry [tries *retries*]
     (when (nil? *session*)
       (throw (RuntimeException.
@@ -149,7 +168,7 @@
                    (with-out-str (pprint (debug-data)))))))
 
     (rc/with-conn [s *session*]
-      (assoc (ssh/ssh s action)
+      (assoc (ssh/ssh s (assoc action :cmd (str/join " " cmd)))
              :host   *host*
              :action action))
     (catch com.jcraft.jsch.JSchException e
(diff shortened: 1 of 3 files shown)


aphyr:
So there's a lot of complexity added here, and I'm not entirely sure the rewards justify adding this to an (already too complicated) part of Jepsen core.

1. Jepsen already supports running without SSH. Add :ssh {:dummy? true}` to your test.
2. There's a new LogSnarf protocol. What is it for? Why is it mixed in with SSH changes?
3. Running commands locally is nice, but a.) doesn't require changing jepsen.control, and b.) as you discovered, changing jepsen.control to support this means layering in a bunch of changes to jepsen.core, which, again, complexity.
4. I know Jepsen is under-tested, but both of these are very core components and are sensitive to changes--I'd like to see comprehensive test suites for old and new behavior so I can have confidence in this kind of rewrite.

May I suggest an alternative path? Perhaps you could use `:dummy?` to disable the SSH machinery--it's built for exactly this sort of use case already. If you want to run shell commands on the local node for startup, you can still put them into `db` but invoke `sh` directly. You can call functions from jepsen.control there to help with escaping and error handling if you like, but keeping the call paths separate will avoid the need for changes to jepsen.core and adding branches to jepsen.control. If there are common functions you'd like to re-use, perhaps you could add a jepsen.control.local namespace to support running local shell commands?

visemet:
> 1. Jepsen already supports running without SSH. Add :ssh {:dummy? true}` to your test.

You're totally right, I hadn't investigated switching to use `:dummy?` after rebasing from Jepsen version 0.1.4. Would you be open to doing something along the lines of [`fake-session`](https://github.com/jepsen-io/jepsen/pull/237/commits/8e77baf65a31a00d9051d1e052d269b61420055d#diff-e4e16e33aa65970d1b55ec53a96ca2a2R514) in order to make the error message more obvious when one of `jepsen.control/ssh*`, `jepsen.control/scp*`, `jepsen.control/upload`, and `jepsen.control/download` are called and `:ssh {:dummy? true}` is being used? I found the `ClassCastException` that comes out of clj-ssh nonobvious when debugging places I had missed using the `jepsen.control/on-local` macro.

> 2. There's a new LogSnarf protocol. What is it for? Why is it mixed in with SSH changes?

I cannot use the `jepsen.core/snarf-logs!` function as written when attempting to "download" the logs from the local node. I don't think `:dummy?` addresses this issue because the `jepsen.core/snarf-logs!` function is going to attempt to call `jepsen.control/download` real session or otherwise. The reason I added the `LogSnarfer` protocol is because I found it counterintuitive that I'd need to make `jepsen.mongodb.core/db` not implement the `LogFiles` protocol in order to make it so that `jepsen.core/snarf-logs!` wouldn't cause an exception when not using SSH. I did end up [copying a fair amount of `jepsen.core.snarf-logs!` when implementing `jepsen.db/snarf-logs!` for `jepsen.mongodb.core/db`](https://github.com/jepsen-io/mongodb/pull/9/commits/4896cc497862e8fbde3b93e7f032a5ec2ad39324#diff-0bac13a42d058dfd6b7ff459e552ca74R315) that I'd be curious if you have a different idea for consolidating these two modes.

> 3. Running commands locally is nice, but a.) doesn't require changing jepsen.control, and b.) as you discovered, changing jepsen.control to support this means layering in a bunch of changes to jepsen.core, which, again, complexity.
>
> If you want to run shell commands on the local node for startup, you can still put them into `db` but invoke `sh` directly. You can call functions from jepsen.control there to help with escaping and error handling if you like, but keeping the call paths separate will avoid the need for changes to jepsen.core and adding branches to jepsen.control. If there are common functions you'd like to re-use, perhaps you could add a jepsen.control.local namespace to support running local shell commands?

The alternative would be to duplicate most of the `jepsen.control/exec` function and at least some of the function it depends on. It wasn't clear to me if the functions inside the `jepsen.control` namespace were intentionally meant to have public visibility, or whether `defn-` just hasn't been judiciously used within Jepsen itself. I'd be fine with defining a `jepsen.control.local` namespace in this repository or putting an implementation of `exec*` via `clojure.java.shell/sh` into the [jepsen-io/mongodb](https://github.com/jepsen-io/mongodb) repository. The former sounds like something that be more useful for others to be able to take advantage of.

aphyr:
> You're totally right, I hadn't investigated switching to use :dummy? after rebasing from Jepsen version 0.1.4. Would you be open to doing something along the lines of fake-session in order to make the error message more obvious when one of jepsen.control/ssh*, jepsen.control/scp*, jepsen.control/upload, and jepsen.control/download are called and :ssh {:dummy? true} is being used? I found the ClassCastException that comes out of clj-ssh nonobvious when debugging places I had missed using the jepsen.control/on-local macro.

Sure! I think you could do it simply by adding a few carefully placed check-and-throws for `*dummy*` to jepsen.control--maybe ssh* or exec? I am 100% in favor of improved error messages and agree we need a better story there.

> The reason I added the LogSnarfer protocol is because I found it counterintuitive that I'd need to make jepsen.mongodb.core/db not implement the LogFiles protocol in order to make it so that jepsen.core/snarf-logs! wouldn't cause an exception when not using SSH.

I don't have time to review the other changes (technically on vacation now ;-)) but perhaps it would be worth having two implementations of DB: one for local, and one for remote nodes? It's totally reasonable to break out shared logic into fns and share them.

To be honest I feel pretty weird about adding this whole other untested execution model to Jepsen for only one test, so... maybe it would make sense to keep as much of this in the mongo test as possible for now, and port things into core once there's demand and the shape of the protocols is worked out. Just thinking as the guy who has to maintain & version this stuff for the next N years. :)

> The alternative would be to duplicate most of the jepsen.control/exec function and at least some of the function it depends on. It wasn't clear to me if the functions inside the jepsen.control namespace were intentionally meant to have public visibility, or whether defn- just hasn't been judiciously used within Jepsen itself. I'd be fine with defining a jepsen.control.local namespace in this repository or putting an implementation of exec* via clojure.java.shell/sh into the jepsen-io/mongodb repository. The former sounds like something that be more useful for others to be able to take advantage of.

If other people ever try to do this, and we can make a compelling case that it isn't something you can do in *current* Jepsen, then I'd say let's add stuff to core. If we can modularize it and keep the branching out of Jepsen core functions, that's also great. My big thing is that this involves a lot of changes to core APIs, and I have a hard time trying to maintain compatibility for Jepsen already--you've been bitten by some semantics changes, yeah? ;-) That's why I try to be as conservative as possible in how we evolve the Jepsen core library. For now, I'd like to see this stuff made as modular as possible, and if that involves some code duplication, I think it's entirely justified. Calling into control internals is better than making control itself more complicated.

visemet:
> Sure! I think you could do it simply by adding a few carefully placed check-and-throws for `*dummy*` to jepsen.control--maybe ssh* or exec? I am 100% in favor of improved error messages and agree we need a better story there.

I'll submit a separate pull request to make that change. Just so it is semi-documented, below is the error message that currently comes out of the functions in the `jepsen.control` namespace when using `:ssh {:dummy? true}`.

```
java.util.concurrent.ExecutionException: java.lang.IllegalArgumentException: No implementation of method: :connected? of protocol: #'clj-ssh.ssh.protocols/Session found for class: clojure.lang.PersistentVector
```

> I don't have time to review the other changes (technically on vacation now ;-)) but perhaps it would be worth having two implementations of DB: one for local, and one for remote nodes? It's totally reasonable to break out shared logic into fns and share them.
>
> To be honest I feel pretty weird about adding this whole other untested execution model to Jepsen for only one test, so... maybe it would make sense to keep as much of this in the mongo test as possible for now, and port things into core once there's demand and the shape of the protocols is worked out.

I've gone ahead and updated jepsen-io/mongodb#9 to define a `jepsen.control.local` namespace per one of your earlier suggestions. I still found it simpler to have only one implemention of the `jepsen.db/DB` protocol with the internals split out between local host and remote host behaviors. Whenever you're back from vacation and can take a closer look, I'd appreciate it.

> Just thinking as the guy who has to maintain & version this stuff for the next N years. :)

Thanks for that, btw :)

----

I'm closing this pull request because `:ssh {:dummy? true}` provides all the functionality I need to be able to avoid using SSH connections.

aphyr:
OK! I actually went to take a stab at doing tests+nice errors for *dummy* mode, and... discovered half an hour in that I don't have the right env set up on this laptop, so it's gonna have to wait until I can get some proper dev time back on the desktop at home. :-(

PR #238, opened 2018-03-13
fixed minor errors in the tutorial

lavacat:
- couple minor typos 
- checker in 06-refining.md didn't work. I've updated with the one from [etcd](https://github.com/jepsen-io/jepsen/blob/master/etcd/src/jepsen/etcd.clj#L161)
- missing final 'etcd-test' update

### DIFF
--- a/doc/tutorial/08-set.md
+++ b/doc/tutorial/08-set.md
@@ -72,7 +72,7 @@ separate keys, or a pool of keys. Another is to use a single key, and have the
 value be a serialized data type, like a JSON array or Clojure set. We'll do the latter.
 
 ```clj
-(defrecord SetClient [conn]
+(defrecord SetClient [k conn]
   client/Client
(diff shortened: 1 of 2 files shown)


doc/tutorial/08-set.md
```diff
@@ -182,7 +182,7 @@ cleaner to have a field on the SetClient.
```
aphyr: Ah, this should have shown the full defrecord--e.g. `(defrecord SetClient [conn k])`. Like the text says, we're gonna add a key parameter.

lavacat: I see, updated with 'k' param

aphyr:
Excellent, thanks Lavacat. This is new stuff I just wrote and haven't had a chance to go through yet, so I really appreciate you sanding off the rough edges like this. :)

PR #241, opened 2018-03-21
Add punctuation to web's test times

mkcp:
Tiny change cause I can't read un-punctuated ISO 8601 dates worth a heck. Also trims the millis at the end which were zeroed out due to `jepsen.util/local-time` so it actually takes up less space!

Before:
screen shot 2018-03-21 at 1 24 43 pm

After:
add punct to web s test times

Tested with etcd


aphyr:
yusssssss

PR #242, opened 2018-03-22
Fix typo: s/commutatative/commutative/ in tutorial

marccarre:
Tata Motors?
Tata, as in, Auntie in French?


aphyr:
It's for cars you can drive on both sides of the road ;-)

PR #243, opened 2018-03-22
Add missing checker dependency in tutorial

marccarre:
Fixes:
```
$ lein run test --nodes-file ../nodes --username admin --time-limit 10 --concurrency 10 -r 1/2
Exception in thread "main" java.lang.RuntimeException: No such namespace: checker, compiling:(jepsen/etcdemo/set.clj:45:13)
    [...]
Caused by: java.lang.RuntimeException: No such namespace: checker
    [...]
```


aphyr:
Thanks Marc!

PR #244, opened 2018-03-24
add java.xml.bind on Java 9

yakovzaytsev:
> lein run test --help

fails in cockroachdb with 

> java.lang.ClassNotFoundException: javax.xml.bind.DatatypeConverter

I had to add java.xml.bind in cockroachdb/project.clj

### DIFF
--- a/cockroachdb/project.clj
+++ b/cockroachdb/project.clj
@@ -8,7 +8,11 @@
                  [org.clojure/java.jdbc "0.6.1"]
                  [circleci/clj-yaml "0.5.5"]
                  [org.postgresql/postgresql "9.4.1211"]]
-  :jvm-opts ["-Xmx12g"
+  :jvm-opts ~(do (require '[clojure.string :as str])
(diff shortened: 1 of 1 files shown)


aphyr:
Hmmmm. I think this means the generated jars will still be broken if they're executed outside lein--what about pulling in the jaxb dependency instead?

yakovzaytsev:
Let me look into it.

tirkarthi:
Can you please attach the full stack trace. I suspect one the dependency is pulling the class that was removed in JDK 9 and could be fixed in the dependency.

tirkarthi:
I haven't tested it but as per https://github.com/reagent-project/reagent-frontend-template/issues/9#issuecomment-374412204 adding a dependency of `[javax.xml.bind/jaxb-api "2.3.0"]` might resolve the issue.

remyrd:
Same behavior on Java 11.
> adding a dependency of [javax.xml.bind/jaxb-api "2.3.0"] might resolve the issue.
It does.

PR #251, opened 2018-04-26
cockroachdb: Add workaround for ntp stop failures

bdarnell:
I've exhausted my ability and patience with debugging this, so just
add a retry.

cockroachdb/cockroach#20492

### DIFF
--- a/cockroachdb/src/jepsen/os/ubuntu.clj
+++ b/cockroachdb/src/jepsen/os/ubuntu.clj
@@ -33,7 +33,9 @@
                   :rsyslog
 		  :tcpdump
                   :logrotate])
-	(c/su (c/exec :service :ntp :stop)))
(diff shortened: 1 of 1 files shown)


aphyr:
Oof, sorry bout that Ben! Yeah, of course we'll merge. :)

bdarnell:
This makes it better, but even with the retry it fails occasionally. Before I try hacking in a sleep or an additional retry (or just ignoring the return value), do you have any idea what might be happening here? It's an exit code of 1 from `systemctl` when stopping ntp. NTP is not running (it was stopped by a previous test and never restarted unless there's some systemd magic going on under the covers. I could never find any evidence of that in the logs).

aphyr:
I've never encountered this one before, so I really don't know! Another option might be to interrogate the status of the service with ps or `service ntp status`, and there's also a process-killing utility in jepsen.control.util that might be helpful, though the service manager might just restart it.

PR #255, opened 2018-05-01
Support SSH sessions to hosts on ports other than 22

rgantt:
I wanted to make it possible to run my clients from a host that cannot directly SSH to my storage nodes (except via a bastion node). So, I set up SSH forwarding on N ports on my local machine to N remote storage nodes. I put each `localhost:port` pair in my `nodes` file.

However, I ran into a problem since `control.clj` was implicitly assuming port 22 (and, specifically, that this port would be the same for all storage nodes).

This change loosens that restriction--specifically, it supports a `nodes` file that contains `hostname:port` pairs. I hope it does so in an unoffensive, backward-compatible way. Being a clojure newbie, I'm confident I've missed some idioms. I also found it very difficult to write a unit test for this (I tried to re-bind `control/clj-ssh-session` in a new `deftest` in `control_tests.clj` but ran into roadblocks and, again, have no idea what I'm doing in clojure).

### DIFF
--- a/jepsen/src/jepsen/control.clj
+++ b/jepsen/src/jepsen/control.clj
@@ -21,7 +21,7 @@
 (def ^:dynamic *sudo*     "User to sudo to"                 nil)
 (def ^:dynamic *username* "Username"                        "root")
 (def ^:dynamic *password* "Password (for login and sudo)"   "root")
-(def ^:dynamic *port*     "SSH listening port"              22)
+(def ^:dynamic *port*     "Default SSH listening port"      22)
 (def ^:dynamic *private-key-path*         "SSH identity file"     nil)
 (def ^:dynamic *strict-host-key-checking* "Verify SSH host keys"  :yes)
 (def ^:dynamic *retries*  "How many times to retry conns"   5)
@@ -267,7 +267,7 @@
 
 (defn clj-ssh-session
   "Opens a raw session to the given host."
-  [host]
+  [host port]
   (let [host  (name host)
         agent (ssh/ssh-agent {})
         _     (when *private-key-path*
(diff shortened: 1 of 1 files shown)


rgantt:
This isn't a particularly interesting feature, but it did enable me to run Jepsen pretty seamlessly on one of our internal databases. Do you have any concerns about merging this? I realize this places another burden on you, but I'm also happy to do whatever it takes to ensure it meshes well with the existing code.

(I'm asking since I have a limited time window to contribute this without further approvals)

aphyr:
Ah, shoot, yes, I looked at this on my phone and forgot to respond! I'm happy to take a PR for configurable ports, but I'd prefer to avoid string munging internally--"host:port" raises other questions, like how do we represent keys, varying usernames etc... I think the right answer here might be to change jepsen.control to take some sort of deeper data structure on a per-node basis. Maybe the :ssh map in tests could be a map of node to {:host, :port, :username, :password, ...}? Naturally we'll have to preserve backwards compatibility with the current SSH tactic, and issue deprecation warnings, but I think those changes can be limited to jepsen.control and jepsen.core without tooo much work; session resolution is pretty much handled by Jepsen as-is. Might cause issues for some tests that assume the dynamic vars for usernames, passwords, etc are in scope though. Hmmm. :/

rgantt:
Thanks! I think that answers both of my questions: am I barking up the wrong tree? and am I barking up the tree correctly?

Let me take another swing at it based on your guidance above. I can worry about extending my approval period as necessary (as long as you don't mind an open PR hanging around).

rgantt:
So, after looking through things a bit more, I would categorize this change as _very_ involved. It requires touching `core`, `control`, and `cli` since the plumbing from the CLI all the way through the test itself has to change. The blast radius is pretty huge, and it needs to be backward compatible, with deprecation warnings, tests, all of that (like you said).

Is there a happy medium here? I totally agree that a fully-correct, per-node configuration is the logical conclusion of a change like this (and I would , but even as-is, it's an improvement which doesn't prevent a future migration (it's imperfect, but not a one-way door).

Would you be comfortable with the change if I factored out the string "munging" into a clearly-documented `fn` instead of doing it in the `let` binding of `session`?

aphyr:
Yeah, I get where you're coming from. What gives me pause is that... once we introduce this feature, folks will start using it, and when we move to per-node configuration later, it'd involve breaking API compatibility, and I try to avoid doing that. Since you're the only person so far who has asked for this, and the patch here is fairly short, I think I'd suggest overriding these functions *in your test code*--you don't even have to fork Jepsen. You can `(ns jepsen.control) (defn ...)` pretty much anywhere to override Jepsen internals.

rgantt:
Awesome, that sounds perfect--and thanks for the clojure pointer. If that works for me, I'm happy! I'll give it a shot and close this PR if it does.

aphyr:
I'm also happy to keep the PR open--I really could be convinced to take this, I'm just leaning against it because the use case seems a little outside the norm. In particular, you want Jepsen to be *as close as possible* to your DB nodes, because every bit of network latency in that path (for instance, having to go through a bastion host) makes it harder for Jepsen to observe ordering violations. But... hey, maybe lots of people want this anyway!

hexiecs:
I have a similar problem. For some reason, the ssh process of my DB nodes can't listen on port 22, so I make them listen on a different port, say 54321. However, it seems that the control node can only connect to port 22, is there any way to work around this problem?

PR #256, opened 2018-05-04
Drop obsolete plugin, update crate tests to jepsen 0.1.8 and fix lost-updates test.

andreidan:
A few minor changes to continue being able to run the tests with the latest CrateDB release (2.3). 

We do not use the sigar plugin anymore so drop that from the crate classpath. 
We update the test suite to use the latest jepsen (0.1.8). 
Fix the lost-updates test to be able to insert the generated `elements` values in the `sets` table 
(to mitigate Lucene's max term length of 32766 bytes).

### DIFF
--- a/crate/src/jepsen/crate/lost_updates.clj
+++ b/crate/src/jepsen/crate/lost_updates.clj
@@ -45,8 +45,8 @@
             (info node "Creating table sets")
             (j/execute! c
                         ["create table if not exists sets (
-                         id     integer primary key,
-                         elements string)"])
+                         id       integer primary key,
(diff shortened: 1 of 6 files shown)


crate/src/jepsen/crate/lost_updates.clj
```diff
@@ -43,8 +45,8 @@
             (info node "Creating table sets")
```
aphyr: You may want to move this code to `setup!`, but I don't think this will break. :)

aphyr:
Looks good, thanks @andreidan!

PR #266, opened 2018-05-17
cockroachdb: Update to head revision of jepsen

bdarnell:
This is a prerequisite for folding the charybdefs nemesis into the main jepsen package (#235). The bulk of the changes here are not strictly necessary but get rid of deprecation warnings.


aphyr:
Looks good, thanks Ben!

Issue #271, opened 2018-06-01
Error during nt/clock-nemesis setup on CentOS


ttyusupov:
```
2018-06-01 20:33:01,888{GMT}    WARN    [jepsen nemesis] jepsen.core: Error setting up nemesis
java.util.concurrent.ExecutionException: java.lang.RuntimeException: sudo -S -u root bash -c "cd /; dpkg --get-selections build-essential" returned non-zero exit status 127 on yb-admin-timur-jepsen-06-01-n1. STDOUT:


STDERR:
bash: dpkg: command not found
```

`yum` should be used instead of `dpkg` on CentOS. I have a fix for that included in this commit - https://github.com/YugaByte/jepsen/commit/6c8a38ffdbcd64a441b66c8b69a5aee5ad3321e7#diff-a311b62d2e79e78232d992fbefa60255R169. I can extract related changed and submit a PR.

ttyusupov:
After fixing "`dpkg` issue", faced another error:
```
ERROR [2018-06-01 20:47:59,371] main - jepsen.cli Oh jeez, I'm sorry, Jepsen broke. Here's why:
java.util.concurrent.ExecutionException: java.lang.RuntimeException: sudo -S -u root bash -c "cd /; ntpdate -b pool.ntp.org" returned non-zero exit status 1 on yb-admin-timur-jepsen-06-01-n1. STDOUT:


STDERR:
 1 Jun 20:47:20 ntpdate[29067]: the NTP socket is in use, exiting
```

Maybe we need to stop NTP daemon at the nemesis start and start again at the nemesis stop?

ttyusupov:
> Maybe we need to stop NTP daemon at the nemesis start and start again at the nemesis stop?
Or maybe even during OS setup/teardown calls since this is OS specific command?

aphyr:
Hey, is this issue for Jepsen proper, or for your fork?

ttyusupov:
I got this issue when using `nt/clock-nemesis` from official `[jepsen "0.1.9"]` with CentOS 7.4 on GCP.

Leviathan1995:
@ttyusupov I find that you are using `[jepsen "0.1.9-SNAPSHOT"]`  not `[jepsen "0.1.9"]`

ttyusupov:
[...]

PR #275, opened 2018-06-04
#271: Overridden build essentials installation for CentOS

ttyusupov:
Had to override build essentials installation for CentOS in order to solve https://github.com/jepsen-io/jepsen/issues/271.

### DIFF
--- a/jepsen/src/jepsen/os.clj
+++ b/jepsen/src/jepsen/os.clj
@@ -5,10 +5,12 @@
   (setup!     [os test node] "Set up the operating system on this particular
                              node.")
   (teardown!  [os test node] "Tear down the operating system on this particular
-                             node."))
+                             node.")
+  (install-build-essential! [os] "Install build tools."))
 
 (def noop
   "Does nothing"
   (reify OS
     (setup!    [os test node])
-    (teardown! [os test node])))
+         (teardown! [os test node])
+         (install-build-essential! [os])))
--- a/jepsen/src/jepsen/os.clj
+++ b/jepsen/src/jepsen/os.clj
@@ -6,11 +6,11 @@
                              node.")
   (teardown!  [os test node] "Tear down the operating system on this particular
                              node.")
-  (install-build-essential! [os] "Install build tools."))
+  (install-build-dependencies! [os] "Install build tools."))
 
 (def noop
   "Does nothing"
   (reify OS
     (setup!    [os test node])
          (teardown! [os test node])
-         (install-build-essential! [os])))
+         (install-build-dependencies! [os])))
--- a/jepsen/project.clj
+++ b/jepsen/project.clj
@@ -1,4 +1,4 @@
-(defproject jepsen "0.1.9"
+(defproject jepsen "0.1.10-SNAPSHOT"
   :description "Call Me Maybe: Network Partitions in Practice"
   :dependencies [[org.clojure/clojure "1.9.0"]
                  [org.clojure/data.fressian "0.2.1"]
--- a/jepsen/src/jepsen/nemesis/time.clj
+++ b/jepsen/src/jepsen/nemesis/time.clj
@@ -46,7 +46,7 @@
     (debian/install [:build-essential])
     (compile-tools!)))
   ([opts]
-   (os/install-build-essential! (:os opts))
+   (os/install-build-dependencies! (:os opts))
    (compile-tools!)))
 
 (defn reset-time!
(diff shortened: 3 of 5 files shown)


jepsen/src/jepsen/os.clj
```diff
@@ -5,10 +5,12 @@
   (setup!     [os test node] "Set up the operating system on this particular
                              node.")
   (teardown!  [os test node] "Tear down the operating system on this particular
-                             node."))
+                             node.")
+  (install-build-essential! [os] "Install build tools."))
```
aphyr: Build-essential is a debian-specific construct; it's not generalizable across other platforms. I... strongly suspect that trying to make compilation truly platform independent will lead to madness, and encourage you to handle this in your test instead.

ttyusupov: Yes, `install-build-essential` sounds confusing, I think it should be renamed to something like `install-build-dependencies` here.

As for handling this in our tests - the thing is that common nemesis `jepsen.nemesis.time` explicitly uses debian-specific function independently of the OS - https://github.com/jepsen-io/jepsen/blob/master/jepsen/src/jepsen/nemesis/time.clj#L40, so to use this nemesis on CentOS I had to do these changes.

And with these changes I was able to run all our tests with common nemeses so far on CentOS.

ttyusupov: Is there some other way to reuse `jepsen.nemesis.time` nemesis under CentOS? Because of explicitly calling debian-specific function on all OSes in https://github.com/jepsen-io/jepsen/blob/master/jepsen/src/jepsen/nemesis/time.clj#L40 this nemesis only works on debian-compatible OSes... Or I am missing something?

aphyr: Yep, like pretty much everything else in Jepsen, it's designed specifically for Debian Jessie. In an ideal world, yeah, every test would run on every operating system, and we'd have zillions of polymorphic functions for doing everything across Debian, Centos, BSDs, Windows, Solaris, etc. Unfortunately, I'm just one person, and I don't have the bandwidth to make that happen.

It's probably best to think of the OS protocol as an escape hatch--it's there when you need to break out of Jepsen's OS-specific assumptions, but adding multi-platform polymorphism like this requires every other OS to play along, and down that path lies suffering. Your best bet is either to test with Debian Jessie, or accept that you're going outside the beaten path, and you'll need to reinvent parts of Jepsen's core in your own OS/nemesis. I try to provide as many modular hooks as possible so that you can do that easily, but no guarantees.

aphyr: Looking at your changes more closely, I think you can do this without too much trouble--you'll make a new nemesis that wraps a jepsen.nemesis.time nemesis, and proxies everything to it, except for the setup! call, where you'll do your centos-specific package installation, then call the installation functions from jepsen.nemesis.time. Maybe 20 lines?

ttyusupov: I think that's right - will try that way, thanks!

However, changes proposed in this PR shouldn't do any harm to existing functionality:
1 - It only adds one function `install-build-dependencies` to OS protocol, which is backward compatible with existing users of OS protocol.
2 - This function is only used from `jepsen.nemesis.time`, so other functionality is not affected.
3 - This change makes `jepsen.nemesis.time` to work under both Debian and CentOS, I agree that this don't fix issues with other OSes, but other OSes won't work with `jepsen.nemesis.time` even in current state, because of debian-specific function call.

Meanwhile, I understand that changes to common code require careful testing of existing functionality and under other OSes to make sure it won't break any existing functionality and you might not have the bandwidth for this...
Maybe it is still worth to keep this PR opened till the time there will be more demand on CentOS support and there will be automated tests for the Jepsen framework itself, so it will quite fast and easy to verify such changes...

aphyr: The idea behind protocols is that they're total; the compiler may allow you to get away with it, but it's just kicking the errors down the road. Adding this not only places an obligation on other OS implementors to implement new functions which they may be unable to satisfy, but also opens the door to an infinite number of multi-os compatibility shims. If you want to make specific improvements to nemesis.time to support both debian and centos go for it, but as the person who has to support this stuff for the indefinite future, making this change at the protocol level is the wrong choice. ;-)

ttyusupov: Is it OK to add a function to `jepsen.os.OS` protocol which will return OS name, so based on it we can decide in `jepsen.nemesis.time` how to install dependencies?

PR #277, opened 2018-06-18
#271: Fixed clock nemesis for CentOS

ttyusupov:
Fix for https://github.com/jepsen-io/jepsen/issues/271.


jepsen/src/jepsen/nemesis/time.clj
```diff
@@ -72,6 +80,9 @@
   (reify nemesis/Nemesis
     (setup! [nem test]
       (c/with-test-nodes test (install!))
```
aphyr: You can make this faster by folding both (install!) and stopping the NTP service inside a single with-test-nodes.

PR #281, opened 2018-07-30
Fix checker function call

dsandubb:
This change updates the RabbitMQ tests in order to be compatible with the new checker queues versions, where the function needs to be invoked beforehand.


aphyr:
Thanks Denisa!

PR #287, opened 2018-09-20
core: Broaden catch when reconnecting

bdarnell:
Most operations are run with blanket `catch Exception` blocks (such as
CockroachDB's `c/with-exception->op`). This one, however, only catches
RuntimeException, any any non-runtime exceptions will escape and be
reported as a test failure. In particular, CockroachDB sees
InterruptedExceptions here with some regularity.

Fixes cockroachdb/cockroach#26279

### DIFF
--- a/jepsen/src/jepsen/core.clj
+++ b/jepsen/src/jepsen/core.clj
@@ -383,7 +383,7 @@
               (try
                 ; Open a new client
                 (set! (.client this) (client/open! (:client test) test node))
-                (catch RuntimeException e
+                (catch Exception e
                   (warn e "Error opening client")
                   (let [fail (assoc op
                                     :type  :fail


bdarnell:
We also see uncaught InterruptedExceptions occasionally in other places (especially when using functions like `gen/delay`, since these sleeps are executed directly without going through any test-specific wrappers). These could perhaps use some catch blocks too, although I'm not sure where exactly they should go or what they should do.

aphyr:
Right. In general I think I've tended to overuse `catch Exception` (which is part of what's gotten us *into* this mess with interrupts), but I think... here it... might be okay. It... might break the way that Jepsen aborts workers when an error is encountered, so... future Jepsen users, if you hit test deadlocks, they might be because interrupts were swallowed here. Let's... cross our fingers and see what happens.

PR #288, opened 2018-09-26
Improvements in Dgraph tests

mkcp:
Bunch of changes in here, mostly in the client and bank tests

- Pull `jepsen.tests.bank/check-op` into a function so it may be called from bank test clients inline

- Bump Dgraph client to version 1.6.0, various changes for compatibility
- Add optional tracing to (almost) all of jepsen.dgraph.client and jepsen.dgraph.bank
- Call `jepsen.tests.bank/check-op` on each bank test read to catch errors as they occur. Useful point to include extra context or halt the test in the future.

### DIFF
--- a/dgraph/src/jepsen/dgraph/core.clj
+++ b/dgraph/src/jepsen/dgraph/core.clj
@@ -112,8 +112,7 @@
    [nil "--replicas COUNT" "How many replicas of data should dgraph store?"
     :default 3
     :parse-fn parse-long
-    ;:validate [pos? "Must be a positive integer"]
-    ]
+    :validate [pos? "Must be a positive integer"]]
    [nil "--rebalance-interval TIME" "How long before automatic rebalances"
     :default "10s"]
    [nil "--final-recovery-time SECONDS" "How long to wait for the cluster to stabilize at the end of a test"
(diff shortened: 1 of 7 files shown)


dgraph/src/jepsen/dgraph/core.clj
```diff
@@ -112,7 +112,8 @@
    [nil "--replicas COUNT" "How many replicas of data should dgraph store?"
     :default 3
     :parse-fn parse-long
-    :validate [pos? "Must be a positive integer"]]
+    ;:validate [pos? "Must be a positive integer"]
```
aphyr: Why remove this?

jepsen/project.clj
```diff
@@ -1,4 +1,4 @@
-(defproject jepsen "0.1.10"
+(defproject jepsen "0.1.10-SNAPSHOT"
```
aphyr: Did you mean 0.1.11-SNAPSHOT?

mkcp:
Got em 

That commented out validation was me playing around w/ weird partition #s. Thanks for catching that and the version

aphyr:
Looks good, thanks Kit!

PR #289, opened 2018-10-01
Set DEBIAN_FRONTEND=noninteractive when installing pacakages

bdarnell:
We occasionally see the error "dpkg-preconfigure: unable to re-open
stdin". I'm not sure why it only happens occasionally, but setting a
more comprehensive non-interactive mode seems like a good idea here.


aphyr:
Thanks Ben!

PR #290, opened 2018-10-01
generator: Catch InterruptedException in delay

bdarnell:
Generator functions are called without any other try/catch
wrapper (unlike most client code which has a blanket catch clause).
gen/delay is interruptible and is a cause of spurious failures during
shutdown (cockroachdb/cockroach#30499).

Note that the approach taken here (proceed immediately when
interrupted) is different from that used in gen/delay-til (which uses
LockSupport/parkNanos for an uninterruptible sleep).

### DIFF
--- a/jepsen/src/jepsen/generator.clj
+++ b/jepsen/src/jepsen/generator.clj
@@ -166,7 +166,10 @@
 (defgenerator DelayFn [f gen]
   [f gen]
   (op [_ test process]
-      (Thread/sleep (* 1000 (f)))
+      (try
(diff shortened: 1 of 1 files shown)


aphyr:
This is a band-aid, but will probably help!

aphyr:
Hey Ben, I'm really sorry bout this, but this patch broke generator time-limit in a bunch of ways. I spent most of the afternoon trying to figure out how to fix it, but wasn't able to sort things out. If you want to take a second shot at this (and get the test suite passing), please be my guest--happy to merge this back in for a future release.

PR #292, opened 2018-10-24
Add clock nemesis to dgraph

mkcp:


aphyr:
Hey Kit, it looks like this duplicates a bunch of logic from jepsen.nemesis.time--do you think it'd make sense to call those functions instead?

mkcp:
omg. I should make more prs just so I can find more parts of jepsen I didn't know of. I took this out of cockroach 😂

yeah i'll update the dgraph tests to use `jepsen.nemesis.time`

mkcp:
Swapped the installation and calls to bump-time over to the `jepsen.nemesis.time` implementations. Not using the nemesis and gens in there cause I'm passing in deltas so the behavior I'm going for is a bit different than the provided ones. Importantly avoids duplicating the installation and logic for calling the script which we definitely do not want a bunch of ways of doing.

aphyr:
Looks good!

PR #294, opened 2018-11-06
Use an HTTPS bintray repository

danielcompton:
This avoids issues where Leiningen won't connect to HTTP repositories.


aphyr:
Thanks Daniel!

PR #295, opened 2018-11-10
make compatiable with dgraph 0.10.0, add tracing features

mkcp:
also finished a todo I left sitting on `with-txn`

### DIFF
--- a/dgraph/src/jepsen/dgraph/trace.clj
+++ b/dgraph/src/jepsen/dgraph/trace.clj
@@ -3,30 +3,48 @@
                                 Tracing
                                 Span)
            (io.opencensus.trace.samplers Samplers)
-           (io.opencensus.exporter.trace.logging LoggingTraceExporter)))
+           (io.opencensus.exporter.trace.logging LoggingTraceExporter)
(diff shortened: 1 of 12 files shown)


dgraph/src/jepsen/dgraph/trace.clj
```diff
@@ -3,30 +3,48 @@
...
-           (io.opencensus.exporter.trace.logging LoggingTraceExporter)))
+           (io.opencensus.exporter.trace.logging LoggingTraceExporter)
+           (io.opencensus.exporter.trace.jaeger JaegerTraceExporter)))
 
-(def tracer ^Tracer (Tracing/getTracer))
-(def trace-exporter (LoggingTraceExporter/register))
+(def tracer         (atom nil))
+(def trace-exporter (atom nil))
+(def sampler        (atom nil))
+(def trace-config   (atom nil))
```
aphyr: I think I missed this earlier, but it feels like something that might belong on the test map, rather than stateful globals? Though if this is just a one-off hack for this test, I don't feel too strongly about it. Thoughts? :)

aphyr:
Cool cool! Thanks Kit!

PR #299, opened 2018-11-27
tablet mover now only sends to leader and quietly times out instead o…

mkcp:
…f throwing

also stripped multiarity from trace attributes


mkcp:
pause on merging please, found some strange behavior with the tablet mover and reviewing if it's these changes or some other issue

Issue #300, opened 2018-11-30
RabbitMQ test issue can't be cast to clojure.lang.IFn


vutruongqa:
I got this issue when running the rabbitmq test: 
The command is : **lein test**
ERROR in (rabbit-test) (rabbitmq_test.clj:56)
Uncaught exception, not in assertion.
expected: nil
  actual: java.lang.ClassCastException: jepsen.checker$reify__7321 cannot be cast to clojure.lang.IFn
 at jepsen.rabbitmq_test/fn (rabbitmq_test.clj:56)
    clojure.test$test_var$fn__7187.invoke (test.clj:704)
    clojure.test$test_var.invoke (test.clj:704)
    clojure.test$test_vars$fn__7209$fn__7214.invoke (test.clj:722)
    clojure.test$default_fixture.invoke (test.clj:674)
    clojure.test$test_vars$fn__7209.invoke (test.clj:722)
    clojure.test$default_fixture.invoke (test.clj:674)
    clojure.test$test_vars.invoke (test.clj:718)
    clojure.test$test_all_vars.invoke (test.clj:728)
    clojure.test$test_ns.invoke (test.clj:747)
    user$eval85$fn__136.invoke (form-init2923833472895342722.clj:1)
    clojure.lang.AFn.applyToHelper (AFn.java:156)
    clojure.lang.AFn.applyTo (AFn.java:144)
    clojure.core$apply.invoke (core.clj:626)
    leiningen.core.injected$compose_hooks$fn__19.doInvoke (form-init2923833472895342722.clj:1)
    clojure.lang.RestFn.applyTo (RestFn.java:137)
    clojure.core$apply.invoke (core.clj:624)
    leiningen.core.injected$run_hooks.invoke (form-init2923833472895342722.clj:1)
    leiningen.core.injected$prepare_for_hooks$fn__24$fn__25.doInvoke (form-init2923833472895342722.clj:1)
    clojure.lang.RestFn.applyTo (RestFn.java:137)
    clojure.lang.AFunction$1.doInvoke (AFunction.java:29)
    clojure.lang.RestFn.invoke (RestFn.java:408)
    clojure.core$map$fn__4245.invoke (core.clj:2559)
    clojure.lang.LazySeq.sval (LazySeq.java:40)
    clojure.lang.LazySeq.seq (LazySeq.java:49)
    clojure.lang.Cons.next (Cons.java:39)
    clojure.lang.RT.boundedLength (RT.java:1654)
[...]

PR #302, opened 2018-12-09
[GH-300] Revert "Fix checker function call"

vjuranek:
Unless jepsen version is also updated (with couple of other
changes implied by jepsen version change), this breaks existing
test.

This reverts commit 2c607aea058c67d41e69f1092f65281fb0af745d.

### DIFF
--- a/rabbitmq/test/jepsen/rabbitmq_test.clj
+++ b/rabbitmq/test/jepsen/rabbitmq_test.clj
@@ -23,8 +23,8 @@
 ;                 :os        debian/os
 ;                 :db        db
 ;                 :client    (mutex)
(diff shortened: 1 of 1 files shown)


aphyr:
Happy to merge this, but some of the other bugs you mentioned may be fixed in newer versions of Jepsen, so you may want to bring these changes back later when you upgrade.

vjuranek:
Sure. Actually I started with updating RabbitMQ test to latest Jepsen version, but I'm getting some strange connection refused exceptions when partitioning happens (which don't pop up with old Jepsen core). This was meant as a quick fix, until I sort it out (which may take a while as I'm playing with it in my free time).

PR #311, opened 2019-02-05
jepsen.checker: remove model arg

mkcp:
Proof of concept using `jepsen.checker/queue` as the test bed. Updated `jepsen.checker-test` to exercise the new behavior and ensure backwards compatible during transition period.


jepsen/src/jepsen/checker.clj
```diff
@@ -63,6 +66,15 @@
          :subdirectory - A directory within this test's store directory where
                          output files should be written. Defaults to nil."))
 
+(defn noop
+  "Creates an empty non-functional checker. 0 arg arity is deprecated."
+  ([]
+   (reify Checker
+     (check [_ _ model _ _])))
+  ([model]
```
aphyr: Since noop doesn't have any concept of a model, we probably don't need an arg here.

jepsen/src/jepsen/checker.clj
```diff
@@ -46,8 +46,11 @@
           true
           valids))
 
+;; DEPRECATED Checkers should be implemented without model. Model should be an
```
aphyr: I'd put this comment in the docstring for `(check ...)`

jepsen/src/jepsen/checker.clj
mkcp: agreed, fixing. found a slightly cleaner way that satisfies the test case

jepsen/src/jepsen/checker.clj
mkcp: done

mkcp:
Example warning:

`11:49:38.486 [nREPL-worker-41] WARN  jepsen.checker - DEPRECATED: `jepsen.checker/check` with arity of 4 not implemented. Falling back to deprecated `jepsen.client/check` with arity of 5. If the checker still needs a model, see `jepsen.checker` documentation for details.`

mkcp:
Anyone using the 5-arity check method will get this when the compiler hits their protocol impl.`Caused by: java.lang.IllegalArgumentException: Can't define method not in interfaces: check`

Which should be good enough to get the line number that has to be changed, and see the note on `jepsen.checker/Checker`. Feel free to @ me for anyone posting questions about it in issues or mailing list ;)

Users will get this message at analysis time if they have a `:model` on their test map. 
`WARN [2019-02-07 20:03:42,987] jepsen test runner - jepsen.core DEPRECATED: Checker model is assigned to test, which is no longer supported. If the checker still needs a model, see `jepsen.checker` documentation for details.`

The highest impact with this change is the linearizable checker, which can be fixed by moving the value of `:model` on test to the checker: `:checker (checker/linearizable {:model (model/cas-register})` Despite the breadth (about 100 references to the old checker and model) making each test and checker compatible is 1 to 2 lines to change at most.

jepsen/src/jepsen/core.clj
```diff
@@ -511,8 +510,7 @@
 
   8. Teardown the operating system
 
-  9. When the generator is finished, invoke the checker with the model and
-     the history
+  9. When the generator is finished, invoke the checker with the history
```
mkcp: update test map docstring

jepsen/src/jepsen/core.clj
```diff
@@ -439,13 +439,13 @@
   (info "Analyzing...")
   (let [; Give each op in the history a monotonically increasing index
         test (assoc test :history (history/index (:history test)))
-
+        _ (when (:model test)
```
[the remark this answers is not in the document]
aphyr: Do you think there might be cases where users want to have their :model on the test map anyway? I don't thiiiink this is something super common, normally it's only there to get folded immediately into the checker, so... this feels OK to me.

jepsen/src/jepsen/tests/causal.clj
```diff
@@ -116,8 +116,7 @@
 (defn cw2 [_ _] {:type :invoke, :f :write, :value 2})
 
 (defn test [opts]
-  {:model (causal-register)
-   :checker (independent/checker (check))
+  {:checker (independent/checker (check (causal-register)))
```
mkcp: I need to redesign parts of the causal checker, it's kind of a mess RIP

aphyr: Is OK!

aphyr:
Kit, this is great work. I think we're good to go here. We should make a note somewhere in the commits that this is a breaking API change, and I'll make sure that gets announced in the changelog for the next release. :)

PR #312, opened 2019-02-07
checker/counter: pre-process history to remove failed ops

mkcp:
Removed both the invocation and completion of failed ops so we didn't wind up with histories missing invocations. It's a touch more expensive, but you never know we might end up composing checkers 🤷‍♀️ 

This will conflict with #311, so let me handle merging if both are approved.

Something to note is that the tests for "interleaved and concurrent reads" and "rolling reads and writes" do not pass with their expected values. I dug into it enough to ensure that this PR didn't affect the behavior. But it's not clear to me if the current behavior or the tested-for behavior is preferable. Worth a look, and I can make changes to the two tests or checker if you want.

### DIFF
--- a/jepsen/src/jepsen/checker.clj
+++ b/jepsen/src/jepsen/checker.clj
@@ -648,14 +648,9 @@
 (defn counter
   "A counter starts at zero; add operations should increment it by that much,
   and reads should return the present value. This checker validates that at
-  each read, the value is at greater than the sum of all :ok increments and
+  each read, the value is greater than the sum of all :ok increments and
   attempted decrements, and lower than the sum of all attempted increments and
   :ok decrements.
-
-  Since we know a :fail increment did not occur, we should decrement the
-  counter by the appropriate amount.
-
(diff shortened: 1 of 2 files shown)


aphyr:
I have not checked the tests carefully, but my intuition is that if they break with this change to remove failed ops, then it's likely the tests (or the checker algorithm) at fault--as far as the algo is concerned, it should be checking a history with all successful or indeterminate ops, and that should be a subset of its correct general-case behavior...

aphyr:
Code looks spot on to me! Good go go! :)

mkcp:
@aphyr I double checked and those two tests are failing on master, no change with this PR or #311. I can take a closer look tomorrow.

aphyr:
Ahhh, that miiight be due to a recent PR I took which re-vamped the counter. I bet they didn't run the tests, and I didn't check the merge commit after.

Hey travis-CI could do this for us, right? That's something to put together at some point!

mkcp:
Yeah I'd be down to set up travis runs on PRs. I'll add it to the plan

PR #313, opened 2019-02-08
CI: add travis.yml and :integration flags for tests

mkcp:
Runs everything _not_ flagged with :integration by default. :integration tests can be run when the control node has access to nodes.


aphyr:
Looks great!

PR #316, opened 2019-02-13
checker: proof of concept test for multiregion nemeses

mkcp:
In this PR we multiarity everything 🕶

### DIFF
--- a/jepsen/test/jepsen/checker_test.clj
+++ b/jepsen/test/jepsen/checker_test.clj
@@ -281,9 +281,13 @@
 
     (testing "can render multiple nemesis regions"
       (let [checker (perf {:nemeses #{{:start #{:start1}
-                                       :stop  #{:stop1}}
+                                       :stop  #{:stop1}
+                                       :fill-color "#800080"
+                                       :transparency 0.1}
                                       {:start #{:start2.1 :start2.2}
-                                       :stop  #{:stop2}}}})
+                                       :stop  #{:stop2.1 :stop2.2}
+                                       :fill-color "#87A96B"
+                                       :transparency 0.1}}})
             test    {:name "nemeses multiregions perf test"
                      :start-time 0}
 
--- a/jepsen/test/jepsen/checker_test.clj
+++ b/jepsen/test/jepsen/checker_test.clj
@@ -3,10 +3,11 @@
   (:use jepsen.checker
         clojure.test)
   (:require [knossos [history :as history]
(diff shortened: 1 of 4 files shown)


aphyr:
Looks on point!

jepsen/test/jepsen/checker_test.clj
```diff
@@ -190,35 +191,79 @@
...
+    ;; TODO Generate nemesis regions
+    (testing "can accept nemesis regions"
+      (let [checker (perf {:nemeses #{{:start #{:start1}
+                                       :stop  #{:stop1}}
+                                      {:start #{:start2.1 :start2.2}
+                                       :stop  #{:stop2}}
+                                      {:start #{:start3}
+                                       :stop  #{:stop3.1 :stop3.2}}
+                                      {:start #{:start4.1 :start4.2 :start4.3}
+                                       :stop  #{:stop4.1 :stop4.2 :stop4.3}}}})
```
mkcp: This is the key section. Do you like this option structure? I'm preeeeeeetty sure I can get it working in the renderer, but I wanted to check in before digging in deeply on this approach.

aphyr: Yeah, I think that'll work just fine. :-)

aphyr: (and it makes sense to do maps with :start and :stop here in case we wanted to let you specify certain colors for nemeses, or maybe shading styles)

mkcp: :D Something just occurred to me: we could be better off using a vector of maps here instead of a set of maps. That way the top->down ordering of the nemesis bars/regions is explicit. Thoughts?

aphyr: Yeah, I'm OK with that too, though generally I have the plotting system sort things on display.

mkcp:
@aphyr out for the long weekend. But if you want to use it, feel free to merge (pending approval)

mkcp:
latency-raw
latency-raw
latency-raw
Some plots from the new tests

mkcp:
Found a bug in the multiregion nemesis opts where the ops in the history didn't miss the specification, so they weren't getting closed. Decided to add some color so you can _actually_  see the multiregion behavior in the test result. (warning, not accessible for red-green colorblindness)

rate

aphyr:
AHHHH SO GOOD

PR #324, opened 2019-02-27
jepsen.tests: port comments test over to causal-reverse

mkcp:
The one outstanding thing to note before we merge is that the `"Can detect reverse causal anomaly"` test case does not pass. So a sequentially inserted read of `[2 1]` is considered valid.

I'm fine with commenting out that test case and revisiting the workload at a later point to expand on its functionality.

### DIFF
--- a/jepsen/test/jepsen/tests/causal_reverse_test.clj
+++ b/jepsen/test/jepsen/tests/causal_reverse_test.clj
@@ -1,0 +1,63 @@
+(ns jepsen.tests.causal-reverse-test
+  (:require [jepsen.tests.causal-reverse :refer :all]
+            [knossos.op :refer :all]
+            [clojure.test :refer :all]
+            [jepsen.checker :as checker]))
+
+(deftest casusal-reverse-test
+  (testing "Can validate sequential histories"
+    (let [c (checker)
+          valid [(invoke 0 :write 1)
+                 (ok     0 :write 1)
+                 (invoke 0 :write 2)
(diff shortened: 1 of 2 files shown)


jepsen/test/jepsen/tests/causal_reverse_test.clj
```diff
@@ -0,0 +1,63 @@
...
+      (is (:valid? (checker/check (checker) nil concurrent2 nil)))))
+
+  (testing "Can detect reverse causal anomaly"
+    (let [c (checker)
+          reverse-causal-read [(invoke 0 :write 1)
+                               (ok     0 :write 1)
+                               (invoke 0 :write 2)
+                               (ok     0 :write 2)
+                               (invoke 0 :read nil)
+                               (ok     0 :read [2 1])]]
```
mkcp: The failing test case

aphyr: Huh, yeah, that does look like a problem! It's been a couple years since I wrote this; maybe I have the test case wrong, or the checker is incorrect.

aphyr: Aha, yes, I misremembered what this checker actually does. We only verify that completed writes appear before later ones, but we don't verify the order they appear in. A valid read has to have both 1 and 2, but in any order. Invalid reads would be [1] or [2], I think.

aphyr:
Thanks Kit :)

PR #325, opened 2019-02-28
Introduce Jepsen tests for the new CP Subsystem module of Hazelcast IMDG

metanet:
CP Subsystem contains linearizable & distributed implementations of the java.util.concurrent APIs offered by Hazelcast. We have been testing it with a new test suite we wrote while developing the new system. We would like to submit these tests to the official Jepsen repo.

Javadoc for the CP Subsystem: https://docs.hazelcast.org/docs/3.12-BETA-1/javadoc/com/hazelcast/cp/CPSubsystem.html

Reference Manual for the CP Subsystem: https://docs.hazelcast.org/docs/3.12-BETA-1/manual/html-single/index.html#cp-subsystem

Our CP subsystem test suite is as follows:

- Non-reentrant lock (`--workload non-reentrant-cp-lock`): In this test case, we test if the new `FencedLock` data structure behaves as a non-reentrant mutex, i.e., it can be held by a single endpoint at a time and only the lock holder endpoint can release it. Moreover, the lock cannot be acquired by the same endpoint reentrantly. It means that while a client holds the lock, it cannot acquire the lock again, without releasing the lock first.

- Reentrant lock (`--workload reentrant-cp-lock`): We test if the new `FencedLock` data structure behaves as a reentrant mutex. The lock instance can be held by a single endpoint at a time and only the lock holder endpoint can release it. Moreover, the current lock holder can reentrantly acquire the lock one more time. Reentrant lock acquire limit is 2 for this test.

- Non-reentrant fenced lock (`--workload non-reentrant-fenced-lock`): `FencedLock` orders lock holders by a monotonic fencing token, which is incremented each time the lock switches from the free state to the held state. In this test case, we validate monotonicity of fencing tokens assigned to subsequent lock holders. Moreover, the lock cannot be acquired by the same endpoint reentrantly. It means that while a client holds the lock, it cannot acquire the lock again, without releasing the lock first.

- Reentrant fenced lock (`--workload reentrant-fenced-lock`): `FencedLock` orders lock holders by a monotonic fencing token, which is incremented each time the lock switches from the free state to the held state. However, if the current lock holder acquires the lock reentrantly, it will get the same fencing token. Reentrant lock acquire limit is 2 for this test.

- Semaphore (`--workload semaphore`): In this test, we initialize our new linearizable `ISemaphore` with 2 permits. Each client acquires and releases a permit in a loop and we validate permits are held by at most 2 clients at a time.

- Unique ID Generation with the new linearizable `IAtomicLong` (`--workload cp-id-gen-long`): In this test, each client generates a unique long id by using a linearizable IAtomicLong instance and we validate uniqueness of generated ids.

- Compare-and-swap Register with the new linearizable `IAtomicLong` (`--workload cp-cas-long`): In this test, clients randomly perform write and compare-and-swap operations.

- Compare-and-swap Register with the new linearizable `IAtomicReference` (`--workload cp-cas-reference`): In this test, clients randomly perform write and compare-and-swap operations.

Please see the README file for more details.

We are happy to answer any questions related to the new impls and tests.

Thanks in advance for your review and happy testing!

Co-authored-by: Mehmet Dogan <mehmet@hazelcast.com>
Co-authored-by: Ensar Basri Kahveci <basri@hazelcast.com>

### DIFF
--- a/hazelcast/repeat_all_cp_tests.sh
+++ b/hazelcast/repeat_all_cp_tests.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+
+if [ $# != 2 ]; then
+	echo "how to use: ./repeat_all_cp_tests.sh repeat test_duration"
+	exit 1
+fi
+
+repeat=$1
+test_duration=$2
+
+run_single_test () {
+    test_name=$1
+    echo "running $test_name test"
+
+    lein run test --workload ${test_name} --time-limit ${test_duration}
+
+    if [ $? != '0' ]; then
+        echo "$test_name test failed"
+        exit 1
+    fi
+}
+
+round="1"
+
+while [ ${round} -le ${repeat} ]; do
+
+    echo "round: $round"
+
+    run_single_test "non-reentrant-cp-lock"
+    run_single_test "reentrant-cp-lock"
+    run_single_test "non-reentrant-fenced-lock"
+    run_single_test "reentrant-fenced-lock"
+    run_single_test "cp-semaphore"
+    run_single_test "cp-id-gen-long"
+    run_single_test "cp-cas-long"
+    run_single_test "cp-cas-reference"
+
+    round=`expr $round \+ 1`
+
+done
--- a/hazelcast/server/src/jepsen/hazelcast_server.clj
+++ b/hazelcast/server/src/jepsen/hazelcast_server.clj
@@ -17,30 +17,28 @@
     :parse-fn (fn [m]
                   (str/split m #"\s*,\s*"))]])
 
-(defn prepareCPSubsystemConfig
+(defn prepare-cp-subsystem-config
   "Prepare Hazelcast CPSubsystemConfig"
   [config members]
   (let [cpSubsystemConfig (.getCPSubsystemConfig config)
         raftAlgorithmConfig (.getRaftAlgorithmConfig cpSubsystemConfig)
         semaphoreConfig (CPSemaphoreConfig. "jepsen.cpSemaphore" false)
         lockConfig1 (FencedLockConfig. "jepsen.cpLock1" 1)
-        lockConfig2 (FencedLockConfig. "jepsen.cpLock2" 2)
+        lockConfig2 (FencedLockConfig. "jepsen.cpLock2" 2)]
 
-        _ (.setLeaderElectionTimeoutInMillis raftAlgorithmConfig 1000)
-        _ (.setLeaderHeartbeatPeriodInMillis raftAlgorithmConfig 1500)
-        _ (.setCommitIndexAdvanceCountToSnapshot raftAlgorithmConfig 250)
-        _ (.setFailOnIndeterminateOperationState cpSubsystemConfig true)
+       (.setLeaderElectionTimeoutInMillis raftAlgorithmConfig 1000)
+       (.setLeaderHeartbeatPeriodInMillis raftAlgorithmConfig 1500)
+       (.setCommitIndexAdvanceCountToSnapshot raftAlgorithmConfig 250)
+       (.setFailOnIndeterminateOperationState cpSubsystemConfig true)
 
-        _ (.setCPMemberCount cpSubsystemConfig (count members))
-        _ (.setSessionHeartbeatIntervalSeconds cpSubsystemConfig 5)
-        _ (.setSessionTimeToLiveSeconds cpSubsystemConfig 300)
+       (.setCPMemberCount cpSubsystemConfig (count members))
+       (.setSessionHeartbeatIntervalSeconds cpSubsystemConfig 5)
+       (.setSessionTimeToLiveSeconds cpSubsystemConfig 300)
 
-        _ (.addSemaphoreConfig cpSubsystemConfig semaphoreConfig)
-        _ (.addLockConfig cpSubsystemConfig lockConfig1)
-        _ (.addLockConfig cpSubsystemConfig lockConfig2)
-
-      ]
-    cpSubsystemConfig))
+       (.addSemaphoreConfig cpSubsystemConfig semaphoreConfig)
+       (.addLockConfig cpSubsystemConfig lockConfig1)
+       (.addLockConfig cpSubsystemConfig lockConfig2)
+       cpSubsystemConfig))
 
 (defn -main
   "Go go go"
@@ -70,7 +68,7 @@
         _       (.setEnabled tcp-ip true)
 
         ; prepare the CP subsystem
-        _ (prepareCPSubsystemConfig config members)
+        _ (prepare-cp-subsystem-config config members)
 
         ; Quorum for split-brain protection
         quorum (doto (QuorumConfig.)
--- a/hazelcast/server/src/jepsen/hazelcast_server.clj
+++ b/hazelcast/server/src/jepsen/hazelcast_server.clj
@@ -5,15 +5,42 @@
             [clojure.java.io :as io]
             [clojure.string :as str])
   (:import (com.hazelcast.core Hazelcast)
+           (com.hazelcast.config.cp FencedLockConfig CPSemaphoreConfig)
            (com.hazelcast.config Config
                                  LockConfig
                                  MapConfig
-                                 QuorumConfig)))
+                                 QuorumConfig)
+           ))
 
 (def opt-spec
   [["-m" "--members MEMBER-LIST" "Comma-separated list of peers to connect to"
     :parse-fn (fn [m]
-                (str/split m #"\s*,\s*"))]])
+                  (str/split m #"\s*,\s*"))]])
+
+(defn prepareCPSubsystemConfig
+  "Prepare Hazelcast CPSubsystemConfig"
+  [config members]
+  (let [cpSubsystemConfig (.getCPSubsystemConfig config)
+        raftAlgorithmConfig (.getRaftAlgorithmConfig cpSubsystemConfig)
+        semaphoreConfig (CPSemaphoreConfig. "jepsen.cpSemaphore" false)
+        lockConfig1 (FencedLockConfig. "jepsen.cpLock1" 1)
+        lockConfig2 (FencedLockConfig. "jepsen.cpLock2" 2)
+
+        _ (.setLeaderElectionTimeoutInMillis raftAlgorithmConfig 1000)
+        _ (.setLeaderHeartbeatPeriodInMillis raftAlgorithmConfig 1500)
+        _ (.setCommitIndexAdvanceCountToSnapshot raftAlgorithmConfig 250)
+        _ (.setFailOnIndeterminateOperationState cpSubsystemConfig true)
+
+        _ (.setCPMemberCount cpSubsystemConfig (count members))
+        _ (.setSessionHeartbeatIntervalSeconds cpSubsystemConfig 5)
+        _ (.setSessionTimeToLiveSeconds cpSubsystemConfig 300)
+
+        _ (.addSemaphoreConfig cpSubsystemConfig semaphoreConfig)
+        _ (.addLockConfig cpSubsystemConfig lockConfig1)
+        _ (.addLockConfig cpSubsystemConfig lockConfig2)
+
+      ]
+    cpSubsystemConfig))
 
 (defn -main
   "Go go go"
@@ -23,24 +50,27 @@
                 summary
                 errors]} (cli/parse-opts args opt-spec)
         config  (Config.)
+        members (:members options)
 
         ; Timeouts
-        _ (.setProperty config "hazelcast.client.heartbeat.interval" "1000")
-        _ (.setProperty config "hazelcast.client.heartbeat.timeout" "5000")
-        _ (.setProperty config "hazelcast.client.invocation.timeout.seconds" "5")
+        _ (.setProperty config "hazelcast.client.max.no.heartbeat.seconds" "90")
         _ (.setProperty config "hazelcast.heartbeat.interval.seconds" "1")
-        _ (.setProperty config "hazelcast.master.confirmation.interval.seconds" "1")
         _ (.setProperty config "hazelcast.max.no.heartbeat.seconds" "5")
-        _ (.setProperty config "hazelcast.max.no.master.confirmation.seconds" "10")
         _ (.setProperty config "hazelcast.operation.call.timeout.millis" "5000")
+        _ (.setProperty config "hazelcast.wait.seconds.before.join" "0")
+        _ (.setProperty config "hazelcast.merge.first.run.delay.seconds" "1")
+        _ (.setProperty config "hazelcast.merge.next.run.delay.seconds" "1")
 
         ; Network config
         _       (.. config getNetworkConfig getJoin getMulticastConfig
                     (setEnabled false))
         tcp-ip  (.. config getNetworkConfig getJoin getTcpIpConfig)
-        _       (doseq [member (:members options)]
+        _       (doseq [member members]
                   (.addMember tcp-ip member))
         _       (.setEnabled tcp-ip true)
+
+        ; prepare the CP subsystem
+        _ (prepareCPSubsystemConfig config members)
 
         ; Quorum for split-brain protection
         quorum (doto (QuorumConfig.)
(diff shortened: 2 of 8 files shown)


hazelcast/server/src/jepsen/hazelcast_server.clj
```diff
@@ -5,15 +5,42 @@
...
+(defn prepareCPSubsystemConfig
+  "Prepare Hazelcast CPSubsystemConfig"
+  [config members]
+  (let [cpSubsystemConfig (.getCPSubsystemConfig config)
+        raftAlgorithmConfig (.getRaftAlgorithmConfig cpSubsystemConfig)
+        semaphoreConfig (CPSemaphoreConfig. "jepsen.cpSemaphore" false)
+        lockConfig1 (FencedLockConfig. "jepsen.cpLock1" 1)
+        lockConfig2 (FencedLockConfig. "jepsen.cpLock2" 2)
+
+        _ (.setLeaderElectionTimeoutInMillis raftAlgorithmConfig 1000)
```
aphyr: Since you're not using any of these return values, and there's no bindings after them, you can move all of these expressions into the body of the `let` expression, and drop the `_`s.

hazelcast/src/jepsen/hazelcast.clj
```diff
@@ -48,38 +50,44 @@
 (def pid-file (str dir "/server.pid"))
 (def log-file (str dir "/server.log"))
 
+(def CLIENT_ID_SEPARATOR "xxxxxxxxxx")
```
aphyr: This feels a bit out of place for Clojure constants; consider (def client-id-separator ...)

hazelcast/src/jepsen/hazelcast.clj
```diff
@@ -48,38 +50,44 @@
...
-        (info err)
-        (info exit)
-        (assert (zero? exit))))))
+    (info "Building server")
+    (let [{:keys [exit out err]} (sh "lein" "uberjar" :dir "server")]
+      (info out)
+      (info err)
+      (info exit)
+      (assert (zero? exit)))))
+ )
```
aphyr: It looks like this might be mis-formatted--did you mean to take this back a level of indentation?

hazelcast/src/jepsen/hazelcast.clj
```diff
@@ -119,37 +127,25 @@
         _ (.setProperty config "hazelcast.client.heartbeat.interval" "1000")
         _ (.setProperty config "hazelcast.client.heartbeat.timeout" "5000")
         _ (.setProperty config "hazelcast.client.invocation.timeout.seconds" "5")
-        _ (.setProperty config "hazelcast.heartbeat.interval.seconds" "1")
-        _ (.setProperty config "hazelcast.master.confirmation.interval.seconds" "1")
-        _ (.setProperty config "hazelcast.max.no.heartbeat.seconds" "5")
-        _ (.setProperty config "hazelcast.max.no.master.confirmation.seconds" "10")
-        _ (.setProperty config "hazelcast.operation.call.timeout.millis" "5000")
```
aphyr: Just as a heads-up, the nemeses were designed with particular timings in mind, which depended (in previous versions) on these parameters; you'll want to make sure you've explored the parameter space and that your nemeses are actually triggering the kinds of failover events you'd like to test.

hazelcast/src/jepsen/hazelcast.clj
```diff
@@ -166,7 +162,83 @@
...
+
+    (invoke! [this test op]
+      (case (:f op)
+        :read (assoc op :type :ok, :value (.get atomic-long))
+        :write (do (.set atomic-long (:value op))
+                   (assoc op :type :ok))
+        :cas (let [[currentV newV] (:value op)]
+               (if (.compareAndSet atomic-long currentV newV)
+                 (assoc op :type :ok)
+                 (assoc op :type :fail :error :cas-failed)
```
aphyr: This isn't super important, but idomatically, you'd close off your parens at the end of this line, instead of scattering them onto later lines. You might want to do a pass through the source to clean these up; not critical, but it'll help you blend in with normal Clojure style. :)

hazelcast/src/jepsen/hazelcast.clj
```diff
@@ -257,55 +329,235 @@
                          gen/once
                          gen/each)})
 
+(defn log-acquire-ok
+  [clientName op]
+  (do
+    (try (Thread/sleep LOGGING_SLEEP_DURATION_MS) (catch Exception _))
```
aphyr: Typically, in Jepsen tests, you want to perform operations as quickly as possible. I'm confused about why you've got all these sleep statements scattered throughout--it might be worth adding some documentation here to explain your approach.

hazelcast/src/jepsen/hazelcast.clj
```diff
@@ -257,55 +329,235 @@
...
+             (case (:f op) :acquire (log-acquire-maybe clientName op e) :release (log-release-maybe clientName op e))
+             (assoc op :type :info, :error :exception)
+             )
+           )
+         )
+       )
+
+     (teardown! [this test]
+       (try
+         (.terminate (.getLifecycleService conn))
```
aphyr: In the other clients, you call `(.shutdown conn)`--do you also want to call `.shutdown` here? Or `(.terminate (.getLifecycleService conn))` in other clients?

hazelcast/src/jepsen/hazelcast.clj
```diff
@@ -257,55 +329,235 @@
...
+  )
+
+(defn cp-semaphore-client
+  ([] (cp-semaphore-client nil nil))
+  ([conn semaphore]
+   (reify client/Client
+     (setup! [_ test node]
+       (let [conn (connect node)
+             sem (.getSemaphore (.getCPSubsystem conn) "jepsen.cpSemaphore")
+             _ (.init sem NUMBER_OF_PERMITS)]
```
aphyr: This can be moved to the body of the let binding.

hazelcast/src/jepsen/hazelcast.clj
```diff
@@ -257,55 +329,235 @@
...
+
+     (invoke! [this test op]
+       (let [clientName (.getName conn)]
+         (try
+           (info (str " " CLIENT_ID_SEPARATOR " " clientName " " CLIENT_ID_SEPARATOR " " op))
+           (case (:f op)
+             :acquire (if (.tryAcquire semaphore 5000 TimeUnit/MILLISECONDS)
+                        (do
+                          (log-acquire-ok clientName op)
+                          (assoc op :type :ok :value {:client clientName :uid (:value op)})
```
aphyr: You use this pattern a lot in this part of the test--`(log-acquire-ok clientname op)`, followed by constructing a completion operation with `(assoc op ...)`. Have you considered having `log-acquire-ok` construct the completion operation for you? Then you could drop the do and assoc throughout these clients; might help shorten things a bit.

hazelcast/src/jepsen/hazelcast.clj
```diff
@@ -257,55 +329,235 @@
...
+           )
+         )
+       )
+
+     (teardown! [this test]
+       (try
+         (.terminate (.getLifecycleService conn))
+         (catch Exception e
+           (warn "exception while client termination")
+           (.printStackTrace e)
```
aphyr: This is a bit worrisome; you generally want Jepsen to abort if you can't tear clients down safely, because it might cause you to leak resources over the course of a test.

hazelcast/src/jepsen/hazelcast.clj
```diff
@@ -257,55 +329,235 @@
...
+  )
+
 (defn lock-client
-  ([] (lock-client nil nil))
-  ([conn lock-name]
+  ([lock-name] (lock-client nil nil lock-name))
+  ([conn lock lock-name]
    (reify client/Client
      (setup! [_ test node]
-       (let [conn (rc/open! (rc/wrapper {:name  "hazelcast"
```
aphyr: I don't exaaactly remember what I was thinking when I wrote this section, but it miiight be the case that we needed reconnects in order to get reliable histories out of Hazelcast's locks. You might want to verify that your histories are actually performing enough operations through the course of a failure and recovery that your checker will be able to identify potential problems. Totally guessing here, and of course the client dynamics may have changed since then. :)

hazelcast/src/jepsen/hazelcast.clj
```diff
@@ -343,22 +595,209 @@
...
+                         (gen/stagger 1/10))
    :final-generator (->> {:type :invoke, :f :read}
                          gen/once
                          gen/each)
-   :checker   (checker/set)})
+   :checker         (checker/set)})
+
+
+
+(defn parseLine [line]
```
aphyr: Functions in Clojure are generally kebab-case. Not a huge deal, this is perfectly legal, just letting you know in case you're looking for style feedback. You may also want to add some docstrings here, because it's not totally clear what format this expects, or what it's intended to do.

hazelcast/src/jepsen/hazelcast.clj
```diff
@@ -343,22 +595,209 @@
...
+  (let
+    [tokens (.split line CLIENT_ID_SEPARATOR)]
+    [(:value (clojure.edn/read-string (nth tokens 2))) (.trim (nth tokens 1))]))
+
+(def clientUidsToClientNames (memoize (fn []
+                            (apply array-map
+                                   (flatten
+                                     (map parseLine
+                                          (.split (:out (sh "grep" CLIENT_ID_SEPARATOR "store/latest/jepsen.log")) "\n")
```
aphyr: I strongly recommend against doing this. Store your client ID mappings in the test map instead.

hazelcast/src/jepsen/hazelcast.clj
```diff
@@ -343,22 +595,209 @@
...
+                                          ))))))
+
+(defn getClient [op]
+  (let [val (:value op)]
+    (if (map? val) (:client val) (get (clientUidsToClientNames) (:value op)))))
+
+(defrecord ReentrantMutex [owner lockCount]
+  Model
+  (step [this op]
+    (let [client (getClient op)]
```
aphyr: I even more recommend against doing this. This is a tight loop; you don't want to be doing IO here, let alone shelling out to grepping a potentially massive logfile.

hazelcast/src/jepsen/hazelcast.clj
```diff
@@ -343,22 +595,209 @@
...
+      (if (nil? client)
+        (do
+          (info "no owner!")
```
aphyr: Uh, ditto, IO here... probably a bad thing! You might be able to get away with this because of knossos' memoization, but, uhhhhhhhhh

hazelcast/src/jepsen/hazelcast.clj
```diff
@@ -343,22 +595,209 @@
...
+                                   (flatten
+                                     (map parseLine
+                                          (.split (:out (sh "grep" CLIENT_ID_SEPARATOR "store/latest/jepsen.log")) "\n")
+                                          ))))))
+
+(defn getClient [op]
+  (let [val (:value op)]
+    (if (map? val) (:client val) (get (clientUidsToClientNames) (:value op)))))
+
+(defrecord ReentrantMutex [owner lockCount]
```
aphyr: OK, so... having read the whole model, I think I understand what you're trying to achieve. I think it might be safer, faster, and simpler for you to use (:process op) as your client ID. It captures the notion of re-entrancy, and means you can get rid of the IO, log grepping, and having to log client IDs in the first place.

hazelcast/src/jepsen/hazelcast.clj
```diff
@@ -343,22 +595,209 @@
...
+(defn createReentrantMutex []
+  "A single reentrant mutex responding to :acquire and :release messages"
+  (ReentrantMutex. nil 0))
+
+
+
+(defrecord OwnerAwareMutex [owner]
+  Model
+  (step [this op]
+    (let [client (getClient op)]
```
aphyr: Ditto here; I think you can get rid of external client IDs and use processes IDs instead.

hazelcast/src/jepsen/hazelcast.clj
```diff
@@ -343,22 +595,209 @@
...
+
+
+(defn getFence [op]
+  (let [val (:value op)]
+    (if (map? val) (:fence val) INVALID_FENCE)))
+
+(defrecord FencedMutex [owner lockFence prevOwner]
+  Model
+  (step [this op]
+    (let [client (getClient op) fence (getFence op)]
```
aphyr: Again here.

hazelcast/src/jepsen/hazelcast.clj
```diff
@@ -343,22 +595,209 @@
...
+  Object
+  (toString [this] (str "owner: " owner " lock fence: " lockFence " prev owner: " prevOwner)))
+
+(defn createFencedMutex []
+  "A fenced mutex responding to :acquire and :release messages and tracking monotonicity of observed fences"
+  (FencedMutex. nil INVALID_FENCE nil))
+
+
+
+(defrecord ReentrantFencedMutex [owner lockCount currentFence highestObservedFence highestObservedFenceOwner]
```
aphyr: hooo boy, OK, you're on your own here haha

hazelcast/src/jepsen/hazelcast.clj
```diff
@@ -374,29 +813,110 @@
...
+                                  :generator (->> [{:type :invoke, :f :acquire}
+                                                   {:type :invoke, :f :release}]
+                                                  cycle
+                                                  gen/seq
+                                                  gen/each
+                                                  (gen/stagger 1/10))
+                                  :checker   (checker/linearizable)
+                                  :model     (model/mutex)}
+   :non-reentrant-cp-lock        {:client    (fenced-lock-client "jepsen.cpLock1")
+                                  :generator (->> [{:type :invoke, :f :acquire :value (.toString (UUID/randomUUID))}
```
aphyr: Did you mean for every acquire to use the same value? Why choose a random value at all?

hazelcast/repeat_all_cp_tests.sh
```diff
@@ -0,0 +1,40 @@
...
+
+    if [ $? != '0' ]; then
+        echo "$test_name test failed"
+        exit 1
+    fi
+}
+
+round="1"
+
+while [ ${round} -le ${repeat} ]; do
```
aphyr: You can also use the built-in `--test-count n` option.

hazelcast/src/jepsen/hazelcast.clj
metanet: When these parameters are set via `ClientConfig`, they are not respected in the server side. So we set them in `hazelcast_server.clj`.

metanet:
> Hey! Happy to merge this, but I've left some comments that might help improve the tests going forward.

Thanks for your review! We will address them and will update the PR accordingly.

hazelcast/src/jepsen/hazelcast.clj
metanet: Yeah. The idea is, we use a side-channel, the Jepsen log, to track which client performed which operation. We parse the log only once and memoize the `uid -> client` mappings. We tried to use `(:process op)` but couldn't do it because `:process` information is not passed to our model while verifying a history. What we understand is, knossos keeps only the `:f` and `:value` fields while doing memoization. Maybe we are missing something here, is there a way to access `(:process op)` in a custom model?

aphyr: That's a good point about memoization, and yes, I think you're right, we do ignore processes there--it's a huge performance boost for most tests, and I've never worked with a datatype that was process-specific before. I still advise against doing this though--I'd store the client mappings in an atom on the test map.

hazelcast/src/jepsen/hazelcast.clj
aphyr: Ah, very good!

metanet:
Hi @aphyr 

Thanks for all your comments. They have been quite useful for improving quality of our test suite. I think we handled all of them now. You can see the description in the second commit for details.

Would you like to have another look before merging the PR? 

Thanks in advance,

aphyr:
Hey! I'm short on time and this kind of review takes a while, but skimming, this looks okay! Thanks for the update. :)

PR #327, opened 2019-03-02
monotonic cycle test

mkcp:
Got it graphing and validating single-register histories right now. It's fast as heck (usually less than 25ms) for every history i've given it, but it stackoverflows right around the 10k read mark, give or take 2k reads. Gonna figure out what's going on there next week.

Also, I would like guidance on how to get this working with read transactions viewing multiple keys at once.

This is benched against a non-cyclic history. Going to start generating and benching non-monotonic histories once I figure out why it's blowing its stack.
screen shot 2019-03-01 at 5 32 50 pm

### DIFF
--- a/jepsen/src/jepsen/tests/monotonic_cycle.clj
+++ b/jepsen/src/jepsen/tests/monotonic_cycle.clj
@@ -1,0 +1,104 @@
+(ns jepsen.tests.monotonic-cycle
+  "A checker which searches for incidents of read skew. Because each register
+  is increment-only, we know that there should never exist a pair of reads r1
+  and r2, such that for two registers x and y, where both registers are
+  observed by both reads, x_r1 < x_r2 and y_r1 > y_r2.
+
+  This problem is equivalent to cycle detection: we have a set of partial
+  orders <x, <y, ..., each of which relates states based on whether x increases
+  or not. We're trying to determine whether these orders are *compatible*.
+
+  Imagine an order <x as a graph over states, and likewise for <y, <z, etc.
+  Take the union of these graphs. If all these orders are compatible, there
+  should be no cycles in this graph.
+
+  To do this, we take each key k, and find all values for k. In general, the
+  ordering relation <k is the transitive closure, but for cycle detection, we
+  don't actually need the full closure--we'll restrict ourselves to k=1's
+  successors being those with k=2 (or, if there are no k=2, use k=3, etc). This
+  gives us a set of directed edges over states for k; we union the graphs for
+  all k together to obtain a graph of all relationships.
+
+  Next, we apply Tarjan's algorithm for strongly connected components, which is
+  linear in edges + vertices (which is why we don't work with the full
+  transitive closure of <k). The existence of any strongly connected components
+  containing more than one vertex implies a cycle in the graph, and that cycle
+  will be within that component.
+
+  This isn't suuuper ideal... the connected component could, I guess, be fairly
+  large, and then it'd be hard to prove where the cycle lies. But this feels
+  like an OK start."
+  (:require [jepsen.checker :as checker]
+            [knossos.op :as op]
+            [clojure.core.reducers :as r]
+            [clojure.set :as set]))
+
+(defn tarjan
+  "Returns the strongly connected components of a graph specified by its nodes
+  and a successor function (next) succs from node to nodes. An implementation of
+  Tarjan's Strongly Connected Components.
+  From: http://clj-me.cgrand.net/2013/03/18/tarjans-strongly-connected-components-algorithm/"
+  [nodes succs]
+  ;; Env is a map from nodes to stack length or nil, nil means the node is
+  ;; known to belong to another SCC (Strongly Connected Component) :stack for
+  ;; the current stack and :sccs for the current set of SCCs.
+  (letfn [(sc [env node]
+            (if (contains? env node)
+              env
+              (let [stack (:stack env)
+                    n     (count stack)
+                    env   (assoc env node n :stack (conj stack node))
+                    env   (reduce (fn [env next]
+                                    (let [env (sc env next)]
+                                      (assoc env node (min (or (env next) n) (env node)))))
+                                  env (succs node))]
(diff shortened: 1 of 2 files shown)


jepsen/src/jepsen/tests/monotonic_cycle.clj
```diff
@@ -0,0 +1,119 @@
...
+  ;; known to belong to another SCC (Strongly Connected Component) :stack for
+  ;; the current stack and :sccs for the current set of SCCs.
+  (letfn [(sc [env node]
+            (if (contains? env node)
+              env
+              (let [stack (:stack env)
+                    n     (count stack)
+                    env   (assoc env node n :stack (conj stack node))
+                    env   (reduce (fn [env next]
+                                    (let [env (sc env next)]
```
mkcp: Tarjan's SCC is a DFS. We recur until we blow the stack out. Maximum depth seems to be around 4000-5000 on my machine. Not sure if this could be lazily evaluated with a loop/recur or some other means.

Also, I don't think this impl is O(e) or O(E + V). And if all optimizations up to that point fail, there are ways to parallelize SCC detection. https://www.cs.vu.nl/~wanf/theses/matei.pdf

aphyr: Yeah, we definitely want to do `(recur)` instead of naive recursion.

aphyr:
> Also, I would like guidance on how to get this working with read transactions viewing multiple keys at once.

Yeah! So my intuition is that reads that only see single keys should always pass, because it's always possible to find some point in time where that read was legal. For multiple keys, I think the graph that we'll wind up searching on will be the union of the graphs with respect to each particular key. So if you have the graph for key `x`, and the graph for key `y`, you should be able to take the union of their edges, and THAT graph might contain cycles. For n keys, we'll just union all n. :)

I don't know whether it's faster to compute the entire graph with respect to `x`, then graph with respect to `y`, and merge them later, or if we just have one big graph and build the whole thing in one pass--I'll leave that to you :)

jepsen/src/jepsen/tests/monotonic_cycle.clj
```diff
@@ -0,0 +1,232 @@
...
+  "Takes a set of component-sets from tarjan's results, identifying if any
+  components are strongly connected (more than 1 element per set)."
+  [components]
+  (let [<=2 (fn [set]
+              {set (<= 2 (count set))})
+        error? (fn [m] (some true? (vals m)))]
+    (->> components
+         (map <=2)
+         (filter error?)
+         (reduce merge))))
```
aphyr: Does this return a bunch of objects like `{set-a true, set-b true, sec-c true, ...}`? Would it make sense to just filter the sets, e.g. `(filter #(< 1 (count %)) components)`?

jepsen/src/jepsen/tests/monotonic_cycle.clj
```diff
@@ -0,0 +1,232 @@
...
+
+(defn valid?
+  "Takes a map of registers to their analyzed components. Returns true if
```
aphyr: I'm a little confused by this, because for a history, we should only be doing doing Tarjan's algorithm once, on a single graph of all operations, rather than once for each register. Is that what's happening here?

aphyr:
Hey Kit! So this is good work, and I really like the Tarjan implementation. I think that you've written a different sort of checker than the one that I was thinking about! I think you could be on to something here, but I'll need to understand more about what exactly the structure of operations is in this checker, and what properties SCC verifies on those operations!

What I was thinking about was a history in which our two operations are something like `{:f :inc :value [:x :y]}`, which means increment the values for registers x and y, and `{:f :read :value {:x 3 :y 5}}`, which means the value of x is presently 3, and y is presently 5. Both reads and writes can operate over any subset of the registers in the test, and each operation can pick a different subset of registers without restriction.

Given a set of reads, we construct a set of precedence graphs <x, <y, ..., where <x relates a read of x = a to all reads observing x = a+1 (or, if a+1 is never observed, a+2, and so on). Given two reads r1 and r2, r1 <x r2 means r2 observed the next higher state of register x. Because all of our operations are *increments*, these partial orders <x, <y, ... are all consistent with the serialization order, assuming the system *is* serializable. If we can show that two orders <x and <y are not compatible with *each other*, e.g. because they imply a cycle, then we can show that no serialization order can possible exist, thereby disproving serializability.

These orders <x, <y, etc have nothing to do with the history order of reads. As far as <x, <y, etc are concerned, a history which sees x=1 before x=0 is equivalent to one which sees x=0 before x=1. (That said, we *could* compute a history order <H or a series of per-process orders like <p1, <p2, ... in *addition* to <x, <y, etc, which can tell us if an individual process, or the system as a whole, observed states moving contradictory to the serialization order. I think that might be related to what you've done in this checker, but I'm not entirely sure yet.

OK, so... <x, <y, etc should all be consistent with each other. Let g be the union of all those per-register graphs, such that r1 g r2 iff r1 <x r2 or r1 <y r2 or .... The per-register graphs <x, <y, ... are compatible if there exist no cycles in g. We apply Tarjan's algorithm to g, obtaining strongly connected components. If there are any components larger than 1, we know that there's a cycle in that each of those components, and the history is nonserializable.

Does that make sense? There's a lot of ways we could compute partial orders over transactions, so this is definitely not the only strategy we could take here. I think your approach might be a different, also valid way of checking some properties of the history, but I'm not exactly sure what they are yet. What do you think?

jepsen/src/jepsen/tests/monotonic_cycle.clj
```diff
@@ -0,0 +1,232 @@
...
+  (loop [graph {}
+         last  {}
+         ;; This will probably have to be a map of keys to last-seen values
+         [op & more] history]
+    (let [val   (:value op)
+          ;; Single register get a key called :register
+          val   (if-not (map? val)
+                  {:register val}
+                  val)
+          pairs (concat val)
```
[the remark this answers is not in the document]
mkcp: You're right, this step can be removed

Issue #309, opened 2019-02-04
Implement jepsen.os.ubuntu


mkcp:
We've seen across various issues (#118, #129, #189, #267, etc.) that users are trying the debian os on ubuntu systems, expecting them to be compatible. They are not!

Kyle and I both use debian, but if you want to use ubuntu or simply want to get involved in jepsen this is a great place to contribute with clear impact. If you want to pick this up, say so and I can help scope out how to do this in jepsen, and provide help with protocols and other clojure features.

See also https://github.com/jepsen-io/jepsen/issues/76

jacobmbr:
Hi, I have a [branch](https://github.com/jepsen-io/jepsen/compare/master...jacobmbr:ubuntu) that runs our internal tests on Ubuntu without issues.

What I did so far:
- Make a minimal `jepsen.os.ubuntu` namespace that resuses `debian` functions for setup.
- Fix breaking deps (`libzip2 -> libzip4`, `iproute -> iproute2`, add `sudo`)
- Cobble together an [Ubuntu Dockerfile](https://github.com/jepsen-io/jepsen/compare/master...jacobmbr:ubuntu#diff-b5c62710589271315456ccae75539cdf) suitable for use with `docker/up.sh`

What I didn't do so far:
- Test the time nemesis, which as far as I can tell is the only usage of debian that is not specific to a test. But I'd expect `(debian/install! [:build-essentials)` to work on a node running Ubuntu.
- Modify existing tests in the `jepsen` repo to use `jepsen.os.ubuntu` to see if they would work.

If this is useful at all I'm happy to do whatever is needed to merge this.

aphyr:
I think we can take a PR for this. There's no need to modify existing Jepsen tests (and I want to mildly encourage folks to work with Debian, because that's the platform which we use internally and can best support, but if you want to write your own tests against the Ubuntu OS, that's just fine. :)

jacobmbr:
Great, thanks for the quick reply. I totally understand defaulting to debian. My tests are running on Debian too, I just saw this issue and thought I'd see if I could pick it up.

I polished it up a little and added a `--ubuntu` flag to the `up.sh` script in PR #331.

mkcp:
Thanks @jacobmbr! If there's anything else on the issues list that you want to pick up, go for it! As always, if you want more info just comment and either Kyle or I will scope it out for you.

PR #331, opened 2019-03-04
Add basic support for Ubuntu in tests and `docker/up.sh`

jacobmbr:
Addresses #309.

The bootstrapping commands for debain and Ubuntu are largely compatible, so this PR just adds a `jepsen.os.ubuntu` namespace that reuses `jepsen.os.debian` commands, substituting the correct package versions for an Ubuntu host.

To test this, I cobbled together a Dockerfile suitable for use with the handy `docker/up.sh` script, so I made that available as well and introduced a `--ubuntu` flag that can be passed to `up.sh`.


aphyr:
Thank you! Just to make my life a little easier when it comes time to write changelogs, can you prefix future commits with a note about what part of Jepsen you're changing? Things like "jepsen.os.ubuntu: add package foo" would be helpful. :)

jacobmbr:
Sure thing. Cheers, and thanks for all of your & Kit's work!

aphyr:
You're welcome! Thanks for your help!

PR #335, opened 2019-03-13
Retry with different host when node removal admin command fails

szegedi:
We experience nemesis sometimes picking an already removed node to perform the removal of another node. 

On those occasions, we get an Clojure exception saying `faunadb-admin` exited with nonzero code, so a cheap way to fix this is to retry with a different node on an exception.

### DIFF
--- a/faunadb/src/jepsen/faunadb/nemesis.clj
+++ b/faunadb/src/jepsen/faunadb/nemesis.clj
@@ -115,14 +115,21 @@
                       (c/on-nodes test [v] (fn [test node]
                                              (auto/kill! test node)
                                              (auto/delete-data-files!)))
-                      (c/on-nodes test [(->> (:nodes topo)
-                                             (map :node)
-                                             (remove #{v}) ; Can't remove self
-                                             vec
-                                             rand-nth)]
-                                  (fn [test local-node]
-                                    @(auto/remove-node! v)
-                                    (info :status (auto/status))))
+                      (util/with-retry []
+                        (c/on-nodes test [(->> (:nodes topo)
+                                               (map :node)
+                                               (remove #{v}) ; Can't remove self
+                                               vec
+                                               rand-nth)]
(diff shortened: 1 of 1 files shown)


szegedi:
FWIW, this PR triggered 3 Travis CI builds, of which 2 completed okay and one failed with a ["sleep interrupted" in a time-limit test](https://travis-ci.com/jepsen-io/jepsen/builds/104290413#L799); an unrelated [other build](https://travis-ci.com/jepsen-io/jepsen/builds/104296657) also failed in the same way around the same time. What I'm trying to say, it looks environmental? I haven't figured out how to retry the build in Travis CI.

aphyr:
OK! So the test error here is a race condition in time-limit, which is... pernicious. No need to worry about that. The fact that the nemesis tries to remove already-removed nodes is... concerning, though. That suggests the code that tries to read the current cluster topology might be inferring it wrong--or it could be a race between when we check the state of the cluster and when we execute node removal. If you have persistent errors with this, you might also want to check on the topology-refreshing code and make sure that it's doing something sensible.

szegedi:
What I experienced wasn't nemesis trying to remove an already removed node, but rather using an already removed node as the agent for removing another node. (Since that also works from the topology, the underlying issue is likely the same, though: a removed node remaining in the topology. So probably not a substantial difference. :-) )

aphyr:
Oh! Hi, good morning, I just woke up and did *not* read carefully enough! I thiiiink https://github.com/fauna/jepsen/blob/9f85b248cbef4f443ad5083708944a39722b86f6/faunadb/src/jepsen/faunadb/nemesis.clj#L119 might be looking at nodes which are a part of the topology but removed or down--you might try using `topo/only-active` here: https://github.com/fauna/jepsen/blob/9f85b248cbef4f443ad5083708944a39722b86f6/faunadb/src/jepsen/faunadb/topology.clj#L89.

PR #341, opened 2019-03-19
Ensure faunadb dependencies are met prior to installing with dpkg.

bitbckt:


aphyr:
This test failure is spurious, no need to worry!

PR #343, opened 2019-03-20
Fix linearizable checker after 7d54a9e.

bitbckt:
I'm not clear whether this is the appropriate change or not. Advice appreciated. :)

### DIFF
--- a/faunadb/src/jepsen/faunadb/register.clj
+++ b/faunadb/src/jepsen/faunadb/register.clj
@@ -70,7 +70,6 @@
   [opts]
   (let [n (count (:nodes opts))]
     {:client    (AtomicClient. nil)
-     :model     (model/cas-register 0)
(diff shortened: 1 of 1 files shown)


faunadb/src/jepsen/faunadb/register.clj
```diff
@@ -82,4 +82,4 @@
      :checker (indy/checker
                 (checker/compose
                   {:timeline     (timeline/html)
-                   :linearizable (checker/linearizable)}))}))
+                   :linearizable (checker/linearizable {:model (model/cas-register)})}))}))
```
aphyr: You'll want to remove :model above, and use (cas-register 0) here. But yes, this is the right idea!

bitbckt:
@aphyr advice implemented.

aphyr:
Thanks! :)

PR #345, opened 2019-03-23
This pull request adds runner, which runs Apache Ignite tests, new bank test and some refactoring.

leapsky:


mkcp:
Great to see you all making progress on these tests! As tests become more complex, it can help to push stuff like DB setup (Kyle's been calling this `jepsen.TEST.auto` for automation) and Client definitions out of the core namespace. Typically the core namespace is called `jepsen.TEST.core` but in this case it looks like `jepsen.ignite` is performing a similar duty. 

I also recommend implementing as much of the test logic as possible in clojure namespaces. Looking over the files for bank.java and client.java, I don't see anything that would stop them from being implemented using java interop.

But, catching the test up to more recent jepsen conventions is not a blocking issue -- I'm happy to merge.

mkcp:
Also, the bank test is implemented as a generic workload in jepsen.tests.bank! I strongly recommend using this implementation so any improvements or fixes we make to it there can be utilized by the ignite test suite

aphyr:
I'll also add that if you find this way of organizing tests (using a `test-fn` map and having each workload (e.g. bank, register) responsible for constructing the entire test) feels a little bit heavyweight, you can clean things up a bit by having workloads construct a partial test map, and having a test function merge that in later. Not at all critical, but it can make it easier to set up complex generator schedules later. See the YugaByte or FaunaDB tests for examples.

PR #353, opened 2019-03-29
Fix line ending errors in docker images built on Windows

alexanderabramov:


mkcp:
Hey, I'm not familiar with dos2unix -- is this compatible on Linux and OS X systems? Don't want to introduce line ending errors elsewhere.

alexanderabramov:
It's used in hundreds of Dockerfiles:
https://github.com/search?l=&q=dos2unix+language%3ADockerfile&type=Code

It just replaces \r\n with \n, so if you are building e.g. on Linux and your files already use \n line endings, it just won't do anything.

mkcp:
OK, thank you for explaining your work.

PR #58, opened 2015-06-02
Docker container support

tjake:
We need a simple way to run Jepsen (with all it's setup) in a single shot.

This creates a docker-in-docker container with all the n1-n5, ssh config, etc setup on container start, so all that's left to-do is to run the test.

PR #356, opened 2019-03-31
tidb: add bank-multitable and g2 test 

cwen0:
In this pr 
* add bank-multitable test to test tidb
* fix some tiny bugs in tidb test


siddontang:
I prefer removing G2 now because we can't support serializable isolation.

maoling:
@siddontang  Could you plz give a help for this [issue](https://github.com/jepsen-io/jepsen/pull/58#issuecomment-478326200). very thanks：D

siddontang:
@maoling 

I have no idea, maybe you meet some network problems? or do you run docker in Mac OS but the memory is limited?

aphyr:
I'd say keep G2 around! It'll be a nice guidepost if you wind up aiming for serializability later. :)

aphyr:
Thanks for these changes; you're all good to go!

siddontang:
Hi @aphyr 

Now TiDB doesn't support serializable isolation and we don't have the plan this year too.

If we support later, we will re-add G2 :-)

PR #362, opened 2019-04-04
jepsen.ignite os dependecy fix 

map7000:
Added the ability to select the OS.
Now you can run tests on an already configured OS and if there are no root rights

Added the ability to select Nemesis

Removed add user code from ignite/src/jepsen/ignite.clj.
ensure-user! does not work on centos because of adduser different syntax. Also does not work if you do not have root rights.

### DIFF
--- a/ignite/src/jepsen/ignite.clj
+++ b/ignite/src/jepsen/ignite.clj
@@ -1,6 +1,7 @@
 (ns jepsen.ignite
   (:require [clojure.tools.logging   :refer :all]
             [clojure.string          :as str]
+            [clojure.java.shell      :as shell]
             [clojure.java.io         :as io]
             [jepsen [core            :as jepsen]
                     [checker         :as checker]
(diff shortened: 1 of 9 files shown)


ignite/src/jepsen/ignite.clj
```diff
@@ -72,17 +72,14 @@
   [version]
   (reify db/DB
     (setup! [_ test node]
+      ;if we have no JDK on nodes, when install it. Works only with Debian/Ubuntu
+      (when-not (re-find #"JAVA_HOME" (get (shell/sh "printenv") :out)) (c/su (debian/install-jdk8!)))
       (info node "Installing Apache Ignite" version)
-      (c/su
-        (debian/install-jdk8!)
-        (cu/ensure-user! user)
```
aphyr: You can parameterize these sorts of calls based on the operating system. See, for example, https://github.com/jepsen-io/jepsen/blob/master/yugabyte/src/yugabyte/auto.clj#L40-L48

aphyr:
Looks good! Looks like there was some kind of issue with the second commit message, but otherwise I'm happy to merge.

map7000:
I will better create new PR. Some git problems :)

PR #363, opened 2019-04-05
jepsen.ignite os dependecy fix

map7000:
Added the ability to select the OS.
Added the ability to select Nemesis.


aphyr:
Good to go, thanks!

PR #364, opened 2019-04-08
tidb: fix wrong args

overvenus:
Fix indeterminate: Wrong number of args (4) passed to: clojure.java.jdbc/query


siddontang:
@overvenus 

CI failed

PR #369, opened 2019-04-09
Remove outdated comment from node Dockerfile

GJL:


mkcp:
Appreciate the thought, but let's keep this one for now. We haven't been running on the de-tutumed (🤔) version for long, so having the link nearby is appropriate. Maybe after another jepsen release or two.

PR #370, opened 2019-04-10
dgraph: bump to Jepsen 0.1.14-SNAPSHOT for long-fork tests.

danielmai:
I'm not sure if it's ideal to use a SNAPSHOT release, but it does mean Dgraph picks up the fix for the long-fork tests.


aphyr:
Hi Daniel! Thanks for checking in, and yes, -SNAPSHOTs should generally be OK--I'm generally using them for my own testing process, so things generally work, but there may be some shifting APIs or transient bugs. 0.1.14-SNAPSHOT introduces a bunch of changes to the way we write graphs, in particular, so you may see some breakage there. Just let me know if you have any issues. :)

PR #371, opened 2019-04-14
 tidb: reduce PD recover time

overvenus:
Tweak a few cfg to reduce PD recover time for efficient testing.


aphyr:
Thank you! :)

PR #372, opened 2019-04-19
Tutorial: update for 0.1.13

siyopao:
Hi, this PR is just some small updates to the tutorial so that it is compatible with Jepsen 0.1.13. Basically moving `:model` from the test opts to the linearizable checker. Thanks!


aphyr:
Great work, thank you!

PR #374, opened 2019-04-19
Docker: use a network insted of links

siyopao:
https://docs.docker.com/network/links/  <--- it is no longer recommended to use links and to use a network instead, so this is what this PR accomplishes. Thank you.


aphyr:
Thank you!

PR #375, opened 2019-04-22
jepsen.os.debian: Support JDK11.

bitbckt:


aphyr:
Are we legally allowed to do this with the changes to the Oracle JDK license agreement? IIRC there's new language about production systems and I'm not sure whether Jepsen would fall under that...

bitbckt:
Hmm. Good point. Perhaps we want OpenJDK11 instead?

aphyr:
I think that might be our safest option, and it'd beat having to do the Oracle installer dance. I know it's a pain, haha.

bitbckt:
It sure is. :) I will update.

bitbckt:
This seems to work well in my testing on a fresh cluster, but be sure Oracle hasn't already poisoned `sources.list` or you're in for a bad time...

aphyr:
I'll take it! Thanks! :)

PR #380, opened 2019-04-27
tidb: require read lock in append workload

overvenus:
`FOR UPDATE` by locking on keys prevents write skew in the append workload.


aphyr:
I've got these changes in master already. Thanks! :)

PR #386, opened 2019-05-09
tidb: add schedule nemesis

overvenus:
Add some special schedulers these are only used in tests, very helpful for testing Multi-Raft implementation.

- shuffle-leader: randomly transfer raft leader between different peers,
- shuffle-region: randomly add peer to different tikv,
- random-merge: randomly merge two regions into one.


aphyr:
Neat, thanks!

PR #387, opened 2019-05-09
tidb: fix create index

cwen0:


aphyr:
Ah, thanks @cwen0! I was just about to dig into that myself. :)

PR #394, opened 2019-06-06
FaunaDB: Update monotonic plot.

bitbckt:
This API has changed significantly, and I can't seem to figure it out in a timely manner... help? :)

### DIFF
--- a/faunadb/src/jepsen/faunadb/monotonic.clj
+++ b/faunadb/src/jepsen/faunadb/monotonic.clj
@@ -269,21 +269,17 @@
                (store/path! test (:subdirectory opts)
                             (str "sequential " filename ".png")))]
     (try
-      (g/raw-plot!
-        (concat (perf/preamble path)
-                (perf/nemesis-regions* history)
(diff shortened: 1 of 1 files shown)


aphyr:
Ah! Yeah, you have two options here. For a really quick fix, you can delete the nemesis-regions code altogether. The plots won't show nemesis activity, though. For a more involved fix (which... should actually make the plot code simpler), use (perf/plot!) and related functions: for an example, see

https://github.com/jepsen-io/jepsen/blob/master/jepsen/src/jepsen/checker/clock.clj#L47-L74

The gist is that you construct a map like {:preamble [gnuplot commands ...], :series [{:title :with :data} ...]; the plot! function then takes care of things like bounds checks, rendering dummy series, nemeses, etc.

bitbckt:
I'm not sure if this works yet - testing now - but I think it's the right idea...

bitbckt:
Got it! :)

aphyr:
Cool cool, thanks!

PR #396, opened 2019-06-11
FaunaDB: Systemd can timeout.

bitbckt:
-_-


aphyr:
Aw man, I was wondering if there might be other codes here! Sorry this bit you. :(

PR #403, opened 2019-07-30
Fix compilation with JDK 12

vjuranek:
JDK 12 removed couple of deprecated JDK option, using these options now results into JDK failure. 
To be able to use JDK 12, also version of http-kit needs to be updated.


aphyr:
Thanks @vjuranek

PR #407, opened 2019-08-26
jepsen.control: Add a `Controller` protocol

DeCarabas:
In my internal environment I can't use SSH with the built-in `clj-ssh`, but most of the functions in the control module are still useful to me. In particular, `jepsen.control.util/install-archive!` and `jepsen.nemesis/partition-random-halves` (among others) are still quite relevant.

Rather than re-implementing all of that stuff, I found it useful to add a `Controller` protocol to `jepsen.control`. The protocol provides a small abstraction over the existing `clj-ssh` code, and every test gets an implementation based on `clj-ssh` by default.

With this protocol in place I was able to implement that protocol myself, and run Jepsen tests on my internal infrastructure.

Does this seem like a useful addition? If so, is this in the right ballpark of implementation?

### DIFF
--- a/jepsen/src/jepsen/control.clj
+++ b/jepsen/src/jepsen/control.clj
@@ -4,7 +4,7 @@
   scripts to open connections to various nodes.
 
   Note that a whole bunch of this namespace refers to things as 'ssh',
-  although they really can apply to any controller, not just SSH."
+  although they really can apply to any remote, not just SSH."
   (:import java.io.File)
   (:require [clj-ssh.ssh    :as ssh]
             [jepsen.util    :as util :refer [real-pmap with-thread-name]]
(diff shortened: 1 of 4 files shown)


DeCarabas:
@aphyr, sorry to bother you, but it looks like jepsen 0.1.15 was released just a little bit after this was merged, but 0.1.15 doesn't contain this change. Is that right? Should the version number in the repo be changed to 0.1.16-SNAPSHOT? Or should 0.1.15 be re-released with this change included?

aphyr:
Yeah, this change felt pretty fundamental, and I didn't have time to test it thoroughly before the 0.1.15 release was due, so I deferred it to the next release. It'll go out in 0.1.16. You're welcome to change the version if you like, or I'll do it the next time I'm working on Jepsen's core.

PR #411, opened 2019-09-06
Update dgraph tests to use the latest version of the Java client.

martinmr:


martinmr:
We made some breaking API changes to the clients. This PR changes the jepsen test to use the latest client.

I tested that tests can be executed successfully:
```
{:latency-graph {:valid? true},
  :rate-graph {:valid? true},
  :valid? true},
 :workload
 {:SI
  {:valid? true,
   :read-count 7515,
   :error-count 0,
   :first-error nil,
   :errors {}},
  :plot {:valid? true},
  :valid? true},
 :valid? true}


Everything looks good! ヽ(‘ー`)ノ
```

aphyr:
Okie, thanks!

PR #413, opened 2019-09-11
support customize ntp-server and front-nodes etc.

ChinaXing:


aphyr:
Again, I'd like to see PRs addressed on one specific issue, or at least a clear rationale as to why I should accept all of these changes together. Right now it's not clear to me why some of these changes exist. That's not saying they're bad, just that you haven't communicated to me, or to future readers of this code, what they're for.

PR #416, opened 2019-09-17
Add options for passing in username and password for wget and cached-wget

sbenderli:
Hello,

We have a use case for downloading packages from a private repository, so it is necessary to pass `--user` and `--pasword` flags to `wget`. I could not find a way to do this currently, and I could not find any examples where authentication is used in the existing examples.

Thanks for this awesome project!

### DIFF
--- a/jepsen/src/jepsen/control/util.clj
+++ b/jepsen/src/jepsen/control/util.clj
@@ -78,15 +78,27 @@
 
 (defn wget!
   "Downloads a string URL and returns the filename as a string. Skips if the
-  file already exists."
+  file already exists.
+
+  Options:
+
+    :force?      Even if we have this cached, download the tarball again anyway.
+    :user?       User for wget authentication. If provided, valid pw must also be provided.
+    :pw?         Password for wget authentication."
   ([url]
-   (wget! url false))
-  ([url force?]
-   (let [filename (.getName (file url))]
-     (when force?
-       (exec :rm :-f filename))
+   (wget! url {:force? false}))
+  ([url opts]
+   (let [filename (.getName (file url))
+         wget-opts std-wget-opts
+         ; second parameter was changed from a boolean flag (force?) to an options map
+         ; this check is here for backwards compatibility
+         opts (if (map? opts) opts {:force? opts})]
+     (when (:force? opts)
+      (exec :rm :-f filename))
+     (when-not (empty? (:user? opts))
+      (concat wget-opts [:--user (:user? opts) :--password (:pw? opts)])) 
      (when (not (exists? filename))
-       (wget-helper! std-wget-opts url))
+      (wget-helper! wget-opts url))
      filename)))
 
 (def wget-cache-dir
@@ -104,12 +116,17 @@
 
   Options:
 
-    :force?     Even if we have this cached, download the tarball again anyway."
+    :force?      Even if we have this cached, download the tarball again anyway.
+    :user?       User for wget authentication. If provided, valid pw must also be provided.
+    :pw?         Password for wget authentication."
   ([url]
-   (wget! url {:force? false}))
+   (cached-wget! url {:force? false}))
   ([url opts]
    (let [encoded-url (String. (b64/encode (.getBytes url)) "UTF-8")
-         dest-file   (str wget-cache-dir "/" encoded-url)]
+         dest-file   (str wget-cache-dir "/" encoded-url)
+         wget-opts   (if (empty? (:user? opts)) 
+                       (concat std-wget-opts [:-O dest-file])
+                       (concat std-wget-opts [:-O dest-file :--user (:user? opts) :--password (:pw? opts)]))]
      (when (:force? opts)
        (info "Clearing cached copy of" url)
        (exec :rm :-rf dest-file))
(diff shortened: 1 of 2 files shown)


jepsen/src/jepsen/control/util.clj
```diff
@@ -80,13 +80,15 @@
   "Downloads a string URL and returns the filename as a string. Skips if the
   file already exists."
   ([url]
-   (wget! url false))
-  ([url force?]
+   (wget! url std-wget-opts false))
+  ([url opts]
+   (wget! url opts false))
+  ([url opts force?]
```
aphyr: Let's make `:force?` an option in the option map here, rather than having a separate argument for it.

jepsen/src/jepsen/control/util.clj
```diff
@@ -102,22 +104,30 @@
...
-  Options:
-
-    :force?     Even if we have this cached, download the tarball again anyway."
+  Params:
+    user        user for wget authentication. If provided, valid pw must also be provided.
+    pw          password for wget authentication.
+    force?      Even if we have this cached, download the tarball again anyway."
   ([url]
-   (wget! url {:force? false}))
-  ([url opts]
```
aphyr: Here, I think it'd make more sense to take an option map rather than having n layers of optional arguments.

sbenderli: Will do!

jepsen/src/jepsen/control/util.clj
sbenderli: Is it a problem that this would break existing usages of `wget` with the force flag?

aphyr: Yeah, I'd check via `(map? opts)` to see whether it's a map being passed in, and dispatch appropriately.

sbenderli: Also, these options are actually what get passed to `wget`. I realize now it's poorly named, so I will rename it to `wget-opts`. This way, `cached-wget` can pass in the `std-wget-opts` plus the additional flags `--user --pw`

jepsen/src/jepsen/control/util.clj
sbenderli: They are now all part of the option map. What do you think?

jepsen/src/jepsen/control/util.clj
```diff
@@ -78,15 +78,29 @@
...
-   (let [filename (.getName (file url))]
-     (when force?
-       (exec :rm :-f filename))
+   (wget! url {:force? false}))
+  ([url opts]
+   (let [filename (.getName (file url))
+         wget-opts std-wget-opts]
+     ; second parameter was changed from a boolean flag (force?) to an options map
+     ; this check is here for backwards compatibility
+     (when (map? opts)
```
sbenderli: @aphyr I added the `map?` check here. `wget` behaves very similarly to `cached-wget`

jepsen/src/jepsen/control/util.clj
```diff
@@ -78,15 +78,29 @@
...
+      (when (:force? opts)
```
aphyr: I feel like it's a bit hard to reason about the duplicated codepaths here. What about something like

```clj
(let [opts (if (map? opts opts {:force? opts})]
  ...)
```

jepsen/src/jepsen/control/util.clj
```diff
@@ -104,20 +118,25 @@
 
   Options:
 
-    :force?     Even if we have this cached, download the tarball again anyway."
+    force?      Even if we have this cached, download the tarball again anyway.
```
aphyr: You might want to keep the leading colons here, just so folks know they're keywords rather than symbols. Not a huge deal. :)

jepsen/src/jepsen/control/util.clj
```diff
@@ -135,8 +154,10 @@
   ([url dest]
    (install-archive! url dest false))
   ([url dest force?]
+   (install-archive! url dest "" "" force?))
+  ([url dest user pw force?]
    (let [local-file (nth (re-find #"file://(.+)" url) 1)
-         file       (or local-file (cached-wget! url {:force? force?}))
```
aphyr: Are we still changing the API for cached-wget! here, or can it take option maps?

jepsen/src/jepsen/control/util.clj
```diff
@@ -104,20 +118,25 @@
...
   ([url opts]
    (let [encoded-url (String. (b64/encode (.getBytes url)) "UTF-8")
-         dest-file   (str wget-cache-dir "/" encoded-url)]
+         dest-file   (str wget-cache-dir "/" encoded-url)
+         wget-opts (concat std-wget-opts [:-O dest-file])]
      (when (:force? opts)
        (info "Clearing cached copy of" url)
        (exec :rm :-rf dest-file))
+     (when-not (empty? (:user? opts))
+       (concat wget-opts [:--user (:user? opts) :--password (:pw? opts)]))
```
aphyr: What... is this supposed to do exactly? It looks like this is a noop. Maybe worth writing integration tests here?

sbenderli: Yeah, I hadn't tested this yet, sorry :( I will try to add some tests here and also run some of the existing tests

aphyr:
This looks workable to me! Want to throw a little integration test on it and we'll merge?

aphyr:
This looks fantastic, thank you! Nice work on the FTP/HTTP integration tests too. :)

PR #417, opened 2019-09-19
FaunaDB: Update for changes to 2.8 and beyond

sprsquish:
* Replica name is now set via the command line when initialising or joining a cluster.
* Initial topology reconciliation has gotten faster, no need to cache.
* Use the newer 2.8 client which does away with AsyncHttpClient and is more considerate of machine resources.


aphyr:
Looks good to me! Thank you. :)

aphyr:
Oh, and if you could please prefix commits changing FaunaDB tests with something like "Fauna: ...", that'd help me when I'm reading through commit logs. Jepsen's a big project these days. :-)

PR #423, opened 2019-10-21
Fix incorrect assert failure message in nemesis/invoke!.

### LINKED ISSUE #422: nemesis/invoke! validation
I have a test in which I'd like to `step` the model when a particular nemesis operation happens.

When I tried the, to me, most obvious thing: returning a map with `:type :ok` in `nemesis/invoke!` I broke an assertion in `jepsen.core`:

```clojure
    ; Validate completion
    (assert (= (:type completion) :info)
            (str "Expected nemesis/invoke! to return a map with :type :ok, :fail, or :info, but received "
                 (pr-str completion) " instead"))
```

It seems to me that either the assert check or the message when the check fails is wrong. I hope it's the former, but I suspect it's the latter? Is there a good reason for only allowing `:info` here?

stevana:
Only :type :info should be allowed, fixes #422.

### DIFF
--- a/jepsen/src/jepsen/core.clj
+++ b/jepsen/src/jepsen/core.clj
@@ -272,7 +272,7 @@
                                                (.getMessage e))))))]
     ; Validate completion
     (assert (= (:type completion) :info)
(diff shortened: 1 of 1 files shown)


aphyr:
Thank you!

PR #436, opened 2020-02-04
jepsen.os.debian: truncate arch while checking installed packages

### LINKED ISSUE #420: "debian/install libzip4" tries to install pkg every time even it's already installed
Currently installed packages are checked by using `dpkg --get-selections pkg-name`. 

```clojure
(defn installed
  "Given a list of debian packages (strings, symbols, keywords, etc), returns
  the set of packages which are installed, as strings."
  [pkgs]
  (let [pkgs (->> pkgs (map name) set)]
    (->> (apply c/exec :dpkg :--get-selections pkgs)
         str/split-lines
         (map (fn [line] (str/split line #"\s+")))
         (filter #(= "install" (second %)))
         (map first)
         set)))
```

But packages which are built for multiple architectures have architecture type appended to the package name. And `libzip4` is one of them.

`dpkg --get-selections libzip4` returns 

```
root@n1:~# dpkg --get-selections libzip4
libzip4:amd64					install
```

That's why every time `debian/install [:libzip4]` called, it's assumed as not-installed. This is not a problem normally but somehow after a few runs, `apt-get install -y --force-yes libzip4` returns with a non-zero status and test fails.

```
WARN [2019-10-10 14:09:08,683] main - jepsen.core Test crashed!
clojure.lang.ExceptionInfo: Command exited with non-zero status -1 on node n4:
sudo -S -u root bash -c "cd /; env DEBIAN_FRONTEND=noninteractive apt-get install -y --force-yes libzip4"
```

mdogan:
Package names reported by `dpkg --get-selections` can contain
architecture at the end of package name, such as `package:amd64`
or `package:i386`.

When `debian/install [:package]` is called, if package name contains
arch, then installed package name does not match and Jepsen tries
to install package on every run.

Fixes #420

### DIFF
--- a/jepsen/src/jepsen/os/debian.clj
+++ b/jepsen/src/jepsen/os/debian.clj
@@ -52,6 +52,7 @@
          (map (fn [line] (str/split line #"\s+")))
          (filter #(= "install" (second %)))
          (map first)
+         (map (fn [p] (str/replace p #":amd64|:i386" {":amd64" "" ":i386" ""})))
          set)))
 
 (defn uninstall!


aphyr:
Thank you!

PR #438, opened 2020-02-15
Create LICENSE

dijonkitchen:


aphyr:
Jepsen already has a license, thank you.

obscurerichard:
See https://github.com/jepsen-io/jepsen/blob/master/jepsen/LICENSE.txt

PR #442, opened 2020-02-25
Add a flag, --logging-json, which enables JSON structured Jepsen logs.

stevana:


aphyr:
Cool! I, uh, should note that Jepsen has a habit of printing some things to stdout as well, so this may not be toootally parseable output. You'll find machine-readable dumps of test runs in `store/latest`. :)

stevana:
Yeah, I noticed. E.g.: https://github.com/jepsen-io/jepsen/blob/f203db2b2715a955337b4b51060599a77ba0f299/jepsen/src/jepsen/control/util.clj#L276 ruins my otherwise nicely structured db logs, :-/.

I'd like to create a unified log of all my db logs and the Jepsen log. Let's see how far I get with `grep`ing my way around the unstructured bits.

aphyr:
If you've got an EDN or Fressian reader handy, you should be able to parse the histories directly from history.edn or test.fressian. That's not gonna include *all* logs, but it does cover operations.

stevana:
Was hoping I could just feed everything to https://github.com/microsoft/logmerge . But thanks, I'll keep that in mind.

aphyr:
Ah. Would you like me to revert?

stevana:
This PR should make it easier to use that tool, so no. :-)

PR #452, opened 2020-04-08
Add wide-area network nemesis to nemesis.combined.

stevana:


jepsen/src/jepsen/nemesis/combined.clj
```diff
@@ -35,6 +45,8 @@
      :minority      - Chooses a random minority of nodes
      :majority      - Chooses a random majority of nodes
      :primaries     - All nodes which we think are primaries
+     :follower      - Chooses a single random node that we don't think is primary
+     :followers     - Chooses a random, non-empty subset of nodes we don't think are primaries
```
aphyr, line 49: For regularity with primaries, I suggest calling this some-secondaries.

jepsen/src/jepsen/nemesis/combined.clj
```diff
@@ -26,6 +27,15 @@
...
 
+(defn- followers
+  "Get all nodes which are followers, i.e. not primaries."
+  [db test]
+  (let [primaries (db/primaries db test)
+        nodes (:nodes test)]
+    (filter
+     (fn [node] (not (some (fn [x] (= node x)) primaries)))
+     nodes)))
+
```
aphyr, line 38: This probably belongs in jepsen.db, and I'd name it "secondaries", for symmetry with "primaries".

jepsen/src/jepsen/nemesis/combined.clj
```diff
@@ -249,6 +263,133 @@
...
                              :color "#A0E9E3"}}})))
 
+(defn wan-nemesis
+  "A nemesis which simulates running the cluster on a wide-area network, e.g.
+  a cluster of nodes split across multiple different AWS regions, i.e. higher
+  average latency, larger latency variance and spottier connectivity.
+
+  This nemesis responds to the following f's:
+
+      :drop     (Drop traffic between two nodes.)
```
aphyr, line 273: Should this be :start-partition?

jepsen/src/jepsen/nemesis/combined.clj
```diff
@@ -249,6 +263,133 @@
...
+                  {:type :info
+                   :f    :drop
+                   :value {:src  src
+                           :dest dest}}))
+        heal {:type :info, :f :heal}
+        drop-heal (gen/flip-flop drop' heal)
+
+        ;; Slow and fast.
+        slow (fn [test _process]
+               (let [mean         50
```
aphyr, line 332: As a later pass, we'll want to parameterize these

jepsen/src/jepsen/nemesis/combined.clj
```diff
@@ -249,6 +263,133 @@
...
+  a cluster of nodes split across multiple different AWS regions, i.e. higher
+  average latency, larger latency variance and spottier connectivity.
+
+  This nemesis responds to the following f's:
+
+      :drop     (Drop traffic between two nodes.)
+      :slow     (Delays network packets.)
+      :flaky    (Introduces randomzied packet loss.)
+      :heal     (End all traffic drops and restore network to fast operation.)
+      :fast     (Removes packet loss and delays.)"
```
aphyr, line 277: Likewise, for all of these, I want to see if we can (where meaningful) preserve symmetry with the other nemesis operations in this nemesis. That might not make sense; I'm not fully sure yet. Might also require expanding the net protocol, which I'm happy to do.

jepsen/src/jepsen/nemesis/combined.clj
```diff
@@ -249,6 +263,133 @@
...
+
+(defn wan-generators
+  "A map with a :generator and :final-generator for WAN-related operations
+  Options are from nemesis-package."
+  [opts]
+  (let [db      (:db opts)
+        faults  (:faults opts)
+        drop?   (contains? faults :drop)
+        slow?   (contains? faults :slow)
+        flaky?  (contains? faults :flaky)
```
aphyr, line 311: So... we've got a bit of a problem here, in that all of these could conceivably apply *either* to individual nodes/pairs of nodes, *or* to datacenters. I'm inclined to hold off on merging this into nemesis.combined until we can figure out how to get those things encoded in a way that makes sense. Maybe datacenters should be a protocol that databases support, and we add datacenter-specific *targets* to the existing partition nemesis?

jepsen/src/jepsen/nemesis/combined.clj
```diff
@@ -249,6 +263,133 @@
...
+        flaky-fast (gen/flip-flop flaky fast)
+
+        modes (cond-> []
+                drop?  (conj drop-heal)
+                slow?  (conj slow-fast)
+                flaky? (conj flaky-fast))
+
+        final (cond-> []
+                drop?             (conj heal)
+                (or slow? flaky?) (conj fast))]
```
aphyr, line 364: These are nice to have, but I think, again, I'd be inclined to split slow/fast, and flaky/perfect, into their own nemesis packages, and make datacenter support a matter of *targets* which could apply to all of them.

jepsen/src/jepsen/nemesis/combined.clj
stevana: Fixed

jepsen/src/jepsen/nemesis/combined.clj
stevana: Fixed.

jepsen/src/jepsen/nemesis/combined.clj
stevana: To be honest, I'm a bit confused about `:drop` and it's relation to the partitions nemesis. I've simply mirrored what's in `jepsen.net`. I guess `:drop` has a direction, i.e. `src` and `dest`?

stevana:
I agree that WAN nemesis should operate on the level of datacenters.

Perhaps I've simply chosen a bad name for the nemesis. What if we call it `bad-network-nemesis`, or something less of a mouthful, instead and leave it targeting individual nodes?

Once you are done figuring out how datacenters work, then it would make sense to lift both `bad-network-nemesis` and `partition-nemesis` to the level of datacentres. Perhaps even some clock skews (bad ntp server?) or kill (power outage?) might make sense to operate on datacenters?

Anyhow, it seems like at least two separate PRs (introduce datacenters and make nemesis.combined datacenter aware) that don't necessarily have to hold up this one from going in (because most of `nemesis.combined` will have to be refactored anyway)? :-)

jepsen/src/jepsen/nemesis/combined.clj
stevana: @Danten pointed out that partitions are also always splitting a cluster in two halves, whereas drop can be between two nodes.

I guess one could argue that that the partition nemesis should be made more general by including what `drop` does? But again this seems like a separate issue, `drop` is currently part of `jepsen.net` and that's what this nemesis uses in its `invoke!`...

aphyr:
> Perhaps I've simply chosen a bad name for the nemesis. What if we call it bad-network-nemesis, or something less of a mouthful, instead and leave it targeting individual nodes?

Naw, it's that (part of) the wan nemesis *already exists* in the form of the partition nemesis package, and it doesn't make sense to have two nemeses that do the same thing. Flaky and slow networks should be new, possibly separate nemesis packages, with basically the same structure.

> (because most of nemesis.combined will have to be refactored anyway)?

I don't think any of the existing nemesis.combined needs to be refactored. I'd start by adding some kind of datacenter target to the existing partition nemesis, relying on a :datacenters field in the test--I think once you've got that, everything else will basically fall into place.

PR #453, opened 2020-04-08
Add ability to control clock nemesis targets.

stevana:
Most of the work here was done by @Danten.

### DIFF
--- a/jepsen/src/jepsen/nemesis/combined.clj
+++ b/jepsen/src/jepsen/nemesis/combined.clj
@@ -234,12 +234,23 @@
                                :strobe-clock          :strobe
                                :bump-clock            :bump}
                               (nt/clock-nemesis)})
-          gen     (->> (nt/clock-gen)
-                       (gen/f-map {:reset          :reset-clock
-                                   :check-offsets  :check-clock-offsets
-                                   :strobe         :strobe-clock
-                                   :bump           :bump-clock})
-                       (gen/delay (:interval opts default-interval)))]
+          db (:db opts)
+          target-specs (:targets (:clock opts) (node-specs db))
+          targets (fn [test] (db-nodes test db
+                                       (some-> target-specs seq rand-nth)))
+          clock-gen (gen/phases
+                     (gen/once {:type :info, :f :check-offsets})
+                     ;; Use a random subset of nodes, specificed by
+                     ;; `targets-specs`, as targets in the generators.
+                     (gen/mix [(reset-gen-select  targets)
+                               (bump-gen-select   targets)
+                               (strobe-gen-select targets)]))
+          gen (->> clock-gen
+                   (gen/f-map {:reset          :reset-clock
+                               :check-offsets  :check-clock-offsets
+                               :strobe         :strobe-clock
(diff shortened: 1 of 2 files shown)


stevana:
Like this?

jepsen/src/jepsen/nemesis/time.clj
```diff
@@ -138,35 +138,50 @@
     (teardown! [_ test]
       (reset-time! test))))
 
-(defn reset-gen
-  "Randomized reset generator. Performs resets on random subsets of the tests'
+(defn reset-gen-select
```
aphyr, line 141: These aren't generators any more; they're functions which return generators, and their use pattern is going to be different. Might want to make that explicit in the docstrings.

jepsen/src/jepsen/nemesis/time.clj
```diff
@@ -138,35 +138,50 @@
...
-   :f     :bump
-   :value (zipmap (util/random-nonempty-subset (:nodes test))
-                  (repeatedly (fn []
-                                (long (* (rand-nth [-1 1])
-                                         (Math/pow 2 (+ 2 (rand 16))))))))})
-
-(defn strobe-gen
-  "Randomized clock strobe generator. On random subsets of the test's nodes,
+  [select]
+  (fn [test process]
```
aphyr, line 145: You could also accomplish this by adding a single-arity form to bump-gen, strobe-gen, etc.

jepsen/src/jepsen/nemesis/combined.clj
```diff
@@ -234,12 +234,20 @@
...
-                                   :strobe         :strobe-clock
-                                   :bump           :bump-clock})
-                       (gen/delay (:interval opts default-interval)))]
+          db (:db opts)
+          target-specs (:targets (:clock opts) (node-specs db))
+          targets (fn [test] (db-nodes test db
+                                       (some-> target-specs seq rand-nth)))
+          clock-gen (gen/phases
+                     (gen/once {:type :info, :f :check-offsets})
+                     (gen/mix ((juxt reset-gen-select bump-gen-select strobe-gen-select)
```
aphyr, line 243: This took me about 15 minutes to figure out, haha. I thought that -select was somehow acting as a verb, but it's actually a function which yields generators. I think I get it now. Comments might be helpful.

jepsen/src/jepsen/nemesis/time.clj
stevana: Fixed.

jepsen/src/jepsen/nemesis/combined.clj
stevana: Simplified and added comment.

jepsen/src/jepsen/nemesis/time.clj
stevana: I thought about doing this, but I figured a separate function would be more clear.

aphyr:
All good, thanks!

aphyr:
Uh, this... doesn't actually compile. Reverting for now.

PR #456, opened 2020-04-24
Stevana/clock targets

stevana:
Second try. Changes since last time:

  * Fixed problem with broken compilation;
  * Updated to use new pure generator api.

I tried testing the code on the `etcd` repo using lein checkouts, but `etcd` doesn't compile anymore because some stuff in `tests.cycle` have been moved to `elle`?

Also sorry about the whitespace changes, I have the habit of hitting `SPC m =` and having CIDER reformat the buffer and it doesn't seem to agree with your formatting style (I thought in Lisp-land code formatting was as solved problem?). Or do you use a different formatter/settings? I'm happy to revert it if it annoys you?

### DIFF
--- a/jepsen/src/jepsen/nemesis/combined.clj
+++ b/jepsen/src/jepsen/nemesis/combined.clj
@@ -15,11 +15,11 @@
   which allow us to start, kill, pause, and resume processes."
   (:require [clojure.tools.logging :refer [info warn]]
             [jepsen [control :as c]
-             [db :as db]
-             [generator :as gen]
-             [nemesis :as n]
-             [util :as util :refer [majority
-                                    random-nonempty-subset]]]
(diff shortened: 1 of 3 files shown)


aphyr:
Yeah, I'd prefer this without the whitespace changes--I work hard on keeping a consistent formatting style! Mostly that's vim-fireplace, but there are a few places (e.g. namespace imports) where I align manually for clarity.

aphyr:
Other than the whitespace, I'd say this looks good to go. Migrating etcd forward shouldn't be a huuuge problem--most of the APIs are the same, just with different function names. I've got client work that I gotta tackle before I can head over and fix that unfortunately. :(

aphyr:
Wooo!

Issue #457, opened 2020-04-29
`com.jcraft.jsch.JSchException: java.lang.NullPointerException` during teardown


stevana:
I've started sporadically seeing `com.jcraft.jsch.JSchException: java.lang.NullPointerException` getting thrown during teardown:

```
...
INFO [2020-04-29 07:35:20,487] jepsen test runner - jepsen.core Snarfing log files
INFO [2020-04-29 07:35:20,608] jepsen node n1 - jepsen.core downloading /opt/db/stdout.log to stdout.log
INFO [2020-04-29 07:35:20,631] jepsen node n2 - jepsen.core downloading /opt/db/stdout.log to stdout.log
INFO [2020-04-29 07:35:20,631] jepsen node n3 - jepsen.core downloading /opt/db/stdout.log to stdout.log
INFO [2020-04-29 07:35:20,640] jepsen node n4 - jepsen.core downloading /opt/db/stdout.log to stdout.log
INFO [2020-04-29 07:35:20,651] jepsen node n5 - jepsen.core downloading /opt/db/stdout.log to stdout.log
INFO [2020-04-29 07:35:22,812] jepsen node n2 - jepsen.control.util Stopping /tmp/db.pid
INFO [2020-04-29 07:35:22,855] jepsen node n5 - jepsen.control.util Stopping /tmp/db.pid
INFO [2020-04-29 07:35:22,860] jepsen node n1 - jepsen.control.util Stopping /tmp/db.pid
INFO [2020-04-29 07:35:22,874] jepsen node n3 - jepsen.control.util Stopping /tmp/db.pid
INFO [2020-04-29 07:35:22,914] jepsen node n4 - jepsen.control.util Stopping /tmp/db.pid
WARN [2020-04-29 07:35:22,944] jepsen node n5 - jepsen.control Encountered error with conn [:control "n5"]; reopening
com.jcraft.jsch.JSchException: java.lang.NullPointerException
	at com.jcraft.jsch.Channel.connect(Channel.java:159) ~[jsch-0.1.53.jar:na]
	at com.jcraft.jsch.Channel.connect(Channel.java:145) ~[jsch-0.1.53.jar:na]
	at clj_ssh.ssh$connect_channel.invokeStatic(ssh.clj:512) ~[jepsen-0.1.18.jar:na]
	at clj_ssh.ssh$connect_channel.invoke(ssh.clj:509) ~[jepsen-0.1.18.jar:na]
[...]

PR #458, opened 2020-04-29
Retry `teardown!`, if it throws exceptions.

stevana:
First go at solving #457

### DIFF
--- a/jepsen/src/jepsen/core.clj
+++ b/jepsen/src/jepsen/core.clj
@@ -177,7 +177,7 @@
        ~@body)
      (finally
        (when-not (:leave-db-running? ~test)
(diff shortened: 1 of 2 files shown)


aphyr:
As I mentioned in #457, I don't think this is the right place to address this issue. I understand adding it for symmetry with setup retries, but setup retries are explicitly present to work around issues in the database itself--e.g., when a vendor's product can't reliably form a cluster. Teardown has so far never been subject to that kind of issue, so I'm not sure if introducing a ::teardown-failed handler is actually worthwhile.

aphyr:
(to be clear, I'm not super opposed to this patch--if it turns out that DB teardowns *are* fragile in the future, this would be a great thing to have!)

PR #460, opened 2020-05-07
Add a docker based instance of `Remote`.

stevana:
This is useful for avoiding the jepsen-control jumpbox container in the
docker-compose based workflow. Instead of `docker exec`ing into the
jepsen-control container and using ssh from there to do stuff to the
nodes, this change lets us `docker exec/cp` directly into the nodes
from our local machine -- thus simplifying the developer workflow.

### DIFF
--- a/jepsen/src/jepsen/control/docker.clj
+++ b/jepsen/src/jepsen/control/docker.clj
@@ -1,0 +1,90 @@
+(ns jepsen.control.docker
+  "The recommended way is to use SSH to setup and teardown databases. It's however
+  sometimes conveniet to be able to setup and teardown the databases using
+  `docker exec` and `docker cp` instead, which is what this namespace helps you
+  do. Use at your own risk, this is an unsupported way of running Jepsen."
+  (:require [clojure.string :as str]
+            [clojure.java.shell :refer [sh]]
+            [slingshot.slingshot :refer [throw+]]
+            [jepsen.control :as c])
+  (:import (jepsen.control Remote)))
+
+(defn resolve-container-id
+  "Takes a host, e.g. `localhost:30404`, and resolves the Docker container id
+  exposing that port. Due to a bug in
+  Docker (https://github.com/moby/moby/pull/40442) this is more difficult than
+  it should be."
+  [host]
+  (if-let [[_address port] (str/split host #":")]
+    (let [ps (:out (sh "docker" "ps"))
+          cid (-> (sh "awk" (str "/" port "/ { print $1 }") :in ps)
+                  :out
+                  str/trim-newline)]
+      (if (re-matches #"[a-z0-9]{12}" cid)
+        cid
+        (throw+ {:type ::invalid-container-id, :container-id cid})))
+    (throw+ {:type ::invalid-host, :host host})))
+
+(defn exec
+  "Execute a shell command on a docker container."
+  [container-id {:keys [cmd] :as opts}]
+  (apply sh
+         "docker" "exec" (c/escape container-id)
+         "sh" "-c" cmd
+         (if-let [in (:in opts)]
+           [:in in]
+           [])))
+
+(defn- path->container
+  [container-id path]
+  (str container-id ":" path))
+
+(defn- unwrap-result
+  "Throws when shell returned with nonzero exit status."
+  [exc-type {:keys [exit] :as result}]
+  (if (zero? exit)
+    result
+    (throw+
+     (assoc result :type exc-type)
+     nil ; cause
+     "Command exited with non-zero status %d:\nSTDOUT:\n%s\n\nSTDERR:\n%s"
+     exit
+     (:out result)
+     (:err result))))
+
+(defn cp-to
+  "Copies files from the host to a container filesystem."
+  [container-id local-paths remote-path]
+  (doseq [local-path (flatten [local-paths])]
+    (->> (sh
+          "docker" "cp"
+          (c/escape local-path)
+          (c/escape (path->container container-id remote-path)))
+         (unwrap-result ::copy-failed))))
+
+(defn cp-from
+  "Copies files from a container filesystem to the host."
+  [container-id remote-paths local-path]
+  (doseq [remote-path (flatten [remote-paths])]
+    (->> (sh
+          "docker" "cp"
+          (c/escape (path->container container-id remote-path))
+          (c/escape local-path))
+         (unwrap-result ::copy-failed))))
+
+(defrecord DockerRemote [container-id]
+  Remote
+  (connect [this host]
+    (assoc this :container-id (resolve-container-id host)))
+  (disconnect! [this]
+    (dissoc this :container-id))
+  (execute! [this action]
+    (exec container-id action))
+  (upload! [this local-paths remote-path _rest]
+    (cp-to container-id local-paths remote-path))
+  (download! [this remote-paths local-path _rest]
+    (cp-from container-id remote-paths local-path)))
+
+(def docker
+  "A remote that does things via `docker exec` and `docker cp`."
+  (->DockerRemote nil))


jepsen/src/jepsen/control.clj
```diff
@@ -360,6 +360,35 @@
 
 (def ssh "A remote that does things via clj-ssh." (SSHRemote. nil))
 
+(defn- undo-command-changes
+  "The `wrap-sudo` and `wrap-cd` functions adds stuff in front of commands which
+  makes docker crap itself, this function undoes that."
+  [cmd]
+  (-> cmd
+      (str/replace-first "sudo -S -u root" "")
+      (str/replace-first "cd /; " "")))
```
aphyr, line 369: Ooh, uhhh, this kind of string munging looks fraught with peril. Maybe it would make more sense to move command wrappers into the Remote protocol?

stevana:
Few questions:

* Shall I wrap all operations in `when-not dummy`, or is that a SSH only thing?
* Can we do without `undo-command-changes`?
* Can we do something better than passing in `nodes` to `docker` and using `.indexOf` to figure out the container name?

stevana:
Build seems broken because `Could not find artifact elle:elle:jar:0.1.1-SNAPSHOT in clojars (https://repo.clojars.org/)`.

jepsen/src/jepsen/control.clj
```diff
@@ -360,6 +360,35 @@
...
+        (let [cmd (str "docker cp " local-path " " (:session this) ":" remote-path)]
+          (sh "sh" "-c" cmd))))
+  (download! [this remote-paths local-path rest]
+    (apply sh "docker" "cp" (str (:session this) ":" remote-paths) local-path rest)))
+
+(defn docker
+  "A remote that does things via `docker exec` and `docker cp`."
+  [nodes]
+  (->DockerRemote nodes nil))
+
```
aphyr, line 391: It looks like this protocol couples hardcoded assumptions in the Jepsen library to particular node choices in the docker config, and... I'm kinda wary about that. To be honest, I'm kinda wary of *any* attempts to make jepsen.control a multi-backend system; this stuff is complicated, involves a lot of baked-in assumptions, and attempts to add local command implementations instead of ssh has had a way of creating tests that nobody else can run. I just ripped a whole bunch of similar code *out* of the Mongo tests this week, because they'd become basically unrunnable in standard Jepsen environments. :-/

So... I'm not shooting this down entirely--I am willing to merge! But I do want to warn you that this is an area that's been historically difficult, and that you might be signing yourself up for a decade of handling support requests. I'm also concerned that you might be using this in everyday development work going forward, and as a result, future patches you (or others!) submit might *not* be tested with normal SSH remotes, and break in truly interesting and time-consuming ways. Are you planning on testing future patches/tests you submit to Jepsen in both SSH and Docker remotes?

aphyr:
Ah, I *just* cut that snapshot yesterday. It's on Clojars now, sorry!

jepsen/src/jepsen/control.clj
stevana: I think I see your concern.

>  I just ripped a whole bunch of similar code out of the Mongo tests this week

Is that change publicly available somewhere so that I can have a look?

> Are you planning on testing future patches/tests you submit to Jepsen in both SSH and Docker remotes?

If somebody make a change unrelated to `Remote`, what is it that can work with one instance of `Remote` but be broken with another? It seems to me that if you have an interface with two implementations that are correct, then you should be able to assume that everything that depend on those interfaces will work the same -- that's the whole point of an interface. Or am I missing something? (Assuming that the hardcoded assumptions are removed.)

jepsen/src/jepsen/control.clj
stevana: I was wondering if I can make `wrap-cd` and `wrap-sudo` not add those strings by using `cd nil` and `sudo nil`, but that's not going to be possible, because that happens before `execute!`, right?

jepsen/src/jepsen/control.clj
aphyr: > It seems to me that if you have an interface with two implementations that are correct, then you should be able to assume that everything that depend on those interfaces will work the same -- that's the whole point of an interface.

I desperately want this to be true, and I think it might mostly be true here, which is why I'm willing to merge! All abstractions are leaky, though, and abstractions around things like "operating systems" or "running commands" tend to be particularly so--when tests have done things like this in the past, they tend to accrue increasing layers of environment-specific dispatch code to work around variances in the environments. MongoDB's test had hundreds of lines devoted to "local vs VM" environment differences. I'm hoping that will be less of an issue here.

For the actual changes, uh... see https://github.com/jepsen-io/mongodb. After n hours of trying to unwind the indirection layers they'd wrapped around jepsen.control, I actually deleted everything and started from scratch. :(

aphyr: tl;dr: I'm one person, working full time, as hard as I can, to make Jepsen a thing. But I can't test or support every combination of platforms, which means there's exactly two "blessed" configurations of Jepsen: [Debian 9] x [SSH remotes] x [iptables] x [LXC, real nodes]. Other implementations of the OS, Remote, Net, protocols, other environments... they're there so you *can* do this kind of polymorphism, but maintaining those implementations is your responsibility, and it can't break the blessed configurations. If you're OK with that, we'll merge (after sorting out the string munging stuff).

jepsen/src/jepsen/control.clj
aphyr: That sounds right, yeah. I'm confused about why this is even a thing we have to do, cuz, like... the sudo and directory management stuff is there for a reason. Undoing it sounds like something that could backfire in terrible ways.

stevana: Some docker images do not seem to have `/` in their filesystem, so `cd /` fails. Likewise with `sudo` most docker images only have one user, `root`, sometimes `sudo` isn't even installed.

aphyr: <sigh> Yeah, that's gonna cause *all kinds* of breakage for tests that set up specific users.

aphyr: (tbh I think the answer here is probably "don't use docker"; it's... almost certainly less work to set up actual VMs or real containers instead of trying to work around behavior like this)

jepsen/src/jepsen/control.clj
keynslug: > It looks like this protocol couples hardcoded assumptions in the Jepsen library to particular node choices in the docker config, and... I'm kinda wary about that.

Sorry for interrupting. FWIW a while ago I made an attempt to overcome this through asking docker daemon for container id with a specific label usually assigned by docker-compose during setup phase.

https://github.com/rbkmoney/riak-jepsenized/blob/fa3cf723dd494da1becb52f9e1d6897d071ca27a/src/jepsenized/riak/control/docker.clj#L27-L30

In retrospect I think I had to name it `DockerComposeRemote` though.

jepsen/src/jepsen/control.clj
stevana: I must have been doing something wrong in the early development of this PR. It doesn't make sense that `cd /` would work over `ssh` but not over `docker exec`, I just tried removing `undo-command-changes` completely and it works. Updated PR.

jepsen/src/jepsen/control.clj
stevana: > I just ripped a whole bunch of similar code out of the Mongo tests this week

> For the actual changes, uh... see https://github.com/jepsen-io/mongodb

That repo hasn't changed in almost two years?

aphyr: Haven't pushed yet.

aphyr:
> Shall I wrap all operations in when-not dummy, or is that a SSH only thing?

I think... yeah, you should probably respect dummy for now. It might be the case that we can get rid of dummy altogether and replace it with a no-op implementation of the Remote protocol--it was invented for cases where people wanted to run tests without *any* SSH connections at all; e.g. testing third-party APIs.

jepsen/src/jepsen/control.clj
stevana: @keynslug Cool, I'll have a look at your solution in detail. Thanks for making me aware of it!

stevana:
* Moved to separate namespace, put a big fat warning saying this isn't the recommend way of running Jepsen

* Stole @keynslug's code to improve escaping and error handling (I hope that's ok with you @keynslug?)

* Removed the need to pass in `nodes` and the ugly `.indexOf` trick to figure out the host name, inspired by @keynslug but ultimately a different approach

* I didn't respect dummy in the end, it seemed enough that the use case you mention "it was invented for cases where people wanted to run tests without any SSH connections at all; e.g. testing third-party APIs." is already covered and it seems unnecessary to support it in two different ways (i.e. via DockerRemote also). I'm happy to try to have a go at the `noop-remote` in a separate PR though.

Thoughts?

aphyr:
Nice work! Yeah, this addresses all my concerns, and I don't see any obvious pitfalls. I also agree, on further thought, that dummy+ssh works fine; no need to support it here. Let's go! :)

PR #464, opened 2020-05-26
Ignite: Parameters refactoring, Bank test rewritten in Clojure

map7000:
Fixed problem with backup parameters
read-from-backup - was not work since could not parse the parameter
backups - now you can set the value to 0

Removed unnecessary function when creating a cache.
getCacheConfiguration

Getting a cache now occurs every time it is accessed, not once when it is created.

Bank test rewritten in Clojure


ignite/src/jepsen/ignite/bank.clj
```diff
@@ -1,122 +1,146 @@
...
-                        (r/filter #(= :read (:f %)))
-                        (r/map (fn [op]
-                          (let [balances (:value op)]
-                            (cond
-                              (not= accounts (count (keys balances)))
-                              {:type :wrong-n
-                               :expected accounts
-                               :found    (count (keys balances))
-                               :op       op}
+      "Balances must all be non-negative and sum to the model's total."
```
aphyr, line 34: By the way, you might be able to save the trouble of writing these checkers by hand by using jepsen.tests.bank. :-)

Issue #465, opened 2020-06-20
Docker script doesn't export JEPSEN_ROOT


vjuranek:
Docker script `up.sh` stopped exporting `JEPSEN_ROOT` env. variable which is required by `docker-compose.dev.yml`. If `JEPSEN_ROOT` is not exported manually, `up.sh --dev` fails with
```
WARNING: The JEPSEN_ROOT variable is not set. Defaulting to a blank string.

[...]

ERROR: for jepsen-control  Cannot create container for service control: create .: volume name is too short, names should be at least two alphanumeric characters

ERROR: for control  Cannot create container for service control: create .: volume name is too short, names should be at least two alphanumeric characters
ERROR: Encountered errors while bringing up the project.
```

PR #466, opened 2020-06-20
docker: export JEPSEN_ROOT again

vjuranek:
Export `JEPSEN_ROOT` again as it's required by `docker-compose.dev.yml`.
Fixes GH #465

### DIFF
--- a/docker/up.sh
+++ b/docker/up.sh
@@ -53,6 +53,7 @@ do
         --dev)
             if [ ! "$JEPSEN_ROOT" ]; then
                 JEPSEN_ROOT="$(cd ../ && pwd)"
+                export JEPSEN_ROOT
                 INFO "JEPSEN_ROOT is not set, defaulting to: $JEPSEN_ROOT"
             fi
             INFO "Running docker-compose with dev config"


aphyr:
@stevana, this undoes some changes from your recent patches--does this look right to you? I'm happy to merge, just wanna make sure this won't break something *else*.

stevana:
Looks good to me.

I believe I didn't notice this problem when testing because I always used to set `JEPSEN_ROOT` explicitly, my bad.

I don't think this will break something else, because 1) it reverts back to how it used to be before my patch, 2) the only place this environment variable is used is in `docker-compose.dev.yml`.

An alternative fix would be to explicitly pass the environment variable to the children processes that need it, e.g. `JEPSEN_ROOT="${JEPSEN_ROOT}" docker-compose -f docker-compose.yml ${COMPOSE} ${DEV} build`, as this is what `export` does. I believe this would be more explicit and hence safe, but because of 2) I think as is done in this PR is fine.

aphyr:
Righto. I'll merge this, and if y'all want to refactor later that's fine!

Issue #474, opened 2020-07-11
Tutorials fails because of lack of java.xml.bind


introom:
The tutorial fails because verschlimmbesserung depends on java.xml.bind,  however later java verions has ditched it.

We may either update verschlimmbesserung or update the project.clj with java.xml.bind, e.g., https://github.com/jepsen-io/jepsen/pull/244

aphyr:
That's a great idea--want to make a PR against verschlimmbesserung?

introom:
Looked into it a little bit, it's not about verschlimmbesserung.  

I bumped the version of jepsen from 0.1.13 (as in the tutorial) to 0.2.0,  the problem disappears.

For jepsen 0.1.13, it fails because it depends on http-kit, which requires j.x.b, cf. https://github.com/http-kit/http-kit/issues/356
```
Caused by: java.lang.ClassNotFoundException: javax.xml.bind.DatatypeConverter
        at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:602)
        at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
        at java.base/java.lang.Class.forName0(Native Method)
        at java.base/java.lang.Class.forName(Class.java:427)
        at clojure.lang.RT.classForName(RT.java:2207)
        at clojure.lang.RT.classForNameNonLoading(RT.java:2220)
        at org.httpkit.server$loading__6706__auto____6135.invoke(server.clj:1)
        at org.httpkit.server__init.load(Unknown Source)
        at org.httpkit.server__init.<clinit>(Unknown Source)
```

In jepsen 0.2.0, while it still depends on http kit 2.3.0, somehow the problem disappears.  Maybe some other dependency has pulled in the j.x.b.

I've created pr to bump the jepsen version in the tutorial.
https://github.com/jepsen-io/jepsen/pull/475

PR #475, opened 2020-07-12
Tutorial: bump jepsen version to 0.2.0

introom:
Bump to 0.2.0 so as to avoid the missing java.xml.bind issue.

cf. https://github.com/jepsen-io/jepsen/issues/474


introom:
for 0.2.0, one has to add pure generator part in the test map..   close the pr for now.

PR #476, opened 2020-07-13
Move simulate and friends from test/ into src/.

### LINKED ISSUE #472: Expose `simulate`, `perfect`, etc from jepsen.generator-test?
I'd like to be able to test my generators inside my project, but I couldn't figure out how to import a namespace from the `test/` directory (`simulate` et al are defined in `jepsen.generator-test` currently), is it even possible?

If not, could we move `simulate`, `perfect`, etc into some namespace that can be imported from other projects?

stevana:
So that other projects can import and use these for testing.

Fixes #472

### DIFF
--- a/jepsen/src/jepsen/generator/test.clj
+++ b/jepsen/src/jepsen/generator/test.clj
@@ -1,8 +1,6 @@
 (ns jepsen.generator.test
   (:require [jepsen.generator :as gen]))
 
-(gen/init!)
-
 (def default-test
   "A default test map."
   {})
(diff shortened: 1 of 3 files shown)


jepsen/src/jepsen/generator/test.clj
```diff
@@ -0,0 +1,175 @@
+(ns jepsen.generator.test
+  (:require [jepsen.generator :as gen]))
+
+(gen/init!)
```
aphyr, line 4: This shouldn't be here--it's gotta be called at runtime, not compile time.

stevana:
Ah I see, ok, fixed!

aphyr:
Cool cool, thanks!

PR #478, opened 2020-07-15
Adding a first simplistic k8s remote, similar to Docker

pavlobaron:
When running the distributed system / store under test on k8s, it's better to use k8s' very own pod-based mechanisms to mess around with "nodes" rather than taking the regular ssh-to-node-or-container based approach.

I took the Docker remote as an example to introduce some very simplistic first k8s remote coming with this PR. I could have isolated this in my own jepsen tests, but thought it might be helpful for other people, and maybe it can grow since k8s is growing, and also all kids of systems get their footprint on it more and more.

Hope it's ok. Thanks.

### DIFF
--- a/jepsen/project.clj
+++ b/jepsen/project.clj
@@ -1,4 +1,4 @@
-(defproject jepsen "0.2.2-SNAPSHOT"
+(defproject jepsen "0.2.1-SNAPSHOT"
   :description "Distributed systems testing framework."
   :url         "https://jepsen.io"
   :license {:name "Eclipse Public License"
--- a/jepsen/project.clj
+++ b/jepsen/project.clj
@@ -1,4 +1,4 @@
-(defproject jepsen "0.2.0"
+(defproject jepsen "0.2.1"
   :description "Distributed systems testing framework."
   :url         "https://jepsen.io"
   :license {:name "Eclipse Public License"
--- a/jepsen/src/jepsen/control/k8s.clj
+++ b/jepsen/src/jepsen/control/k8s.clj
@@ -107,5 +107,4 @@
              (str "kubectl get pods " context " " namespace
                   " | tail -n +2 "
                   " | awk '{print $1}'"))]
-    (info (str "" res))
     (lazy-seq (split-lines (trim (:out res))))))
--- a/jepsen/src/jepsen/control/k8s.clj
+++ b/jepsen/src/jepsen/control/k8s.clj
@@ -92,8 +92,9 @@
   (download! [this remote-paths local-path _rest]
     (cp-from context namespace (:pod-name this) remote-paths local-path)))
 
-(def k8s
-  "A remote that does things via `kubectl exec` and `kubectl cp`."
+(defn k8s
+  "Returns a remote that does things via `kubectl exec` and `kubectl cp`, in the default context and namespacd."
+  []
   (->K8sRemote nil nil))
 
 (defn list-pods
--- a/jepsen/src/jepsen/control/k8s.clj
+++ b/jepsen/src/jepsen/control/k8s.clj
@@ -79,7 +79,6 @@
 (defrecord K8sRemote [context namespace]
   Remote
   (connect [this pod-name]
-    "Ignore host assuming direct kubectl cli access. Host parameter is misused to carry the name of the pod"
     (assoc this
            :context (or-parameter "context" context)
            :namespace (or-parameter "namespace" namespace)
--- a/jepsen/src/jepsen/control/k8s.clj
+++ b/jepsen/src/jepsen/control/k8s.clj
@@ -1,0 +1,112 @@
+(ns jepsen.control.k8s
+  "The recommended way is to use SSH to setup and teardown databases.
+  It's however sometimes conveniet to be able to setup and teardown
+  the databases using `kubectl` instead, which is what this namespace
+  helps you do. Use at your own risk, this is an unsupported way
+  of running Jepsen."
+  (:require [clojure.java.shell :refer [sh]]
+            [slingshot.slingshot :refer [throw+]]
+            [jepsen.control :as c]
+            [clojure.string :refer [split-lines, trim]]
+            [clojure.tools.logging :refer [info]])
+  (:import (jepsen.control Remote)))
+
+(defn exec
+  "Execute a shell command on a pod."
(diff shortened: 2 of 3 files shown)


jepsen/src/jepsen/control/k8s.clj
```diff
@@ -0,0 +1,111 @@
...
+  (apply sh
+         "kubectl"
+         "exec"
+         (c/escape pod-name)
+         context
+         namespace
+         "--"
+         "sh"
+         "-c"
+         cmd
```
aphyr, line 26: I don't remember sh's escaping semantics off the top of my head, and I don't know kubectl's at all, so I'd want to be really cautious here about weird inputs--it'd be worth testing with shell magic characters--quotes, semicolons, ampersands, that kind of thing.

jepsen/src/jepsen/control/k8s.clj
```diff
@@ -0,0 +1,111 @@
...
+  (disconnect! [this]
+    (dissoc this :context :namespace :pod-name))
+  (execute! [this action]
+    (exec context namespace (:pod-name this) action))
+  (upload! [this local-paths remote-path _rest]
+    (cp-to context namespace (:pod-name this) local-paths remote-path))
+  (download! [this remote-paths local-path _rest]
+    (cp-from context namespace (:pod-name this) remote-paths local-path)))
+
+(def k8s
```
aphyr, line 95: I think it might be best to make this a function rather than a constant.

jepsen/src/jepsen/control/k8s.clj
```diff
@@ -0,0 +1,111 @@
...
+  [context namespace]
+  (let [context (or-parameter "context" context)
+        namespace (or-parameter "namespace" namespace)
+        res (sh
+             "sh"
+             "-c"
+             (str "kubectl get pods " context " " namespace
+                  " | tail -n +2 "
+                  " | awk '{print $1}'"))]
+    (info (str "" res))
```
aphyr, line 110: You can write `(str res)` here instead. Or just `res`, I think? Do you want this to log normally?

jepsen/src/jepsen/control/k8s.clj
```diff
@@ -0,0 +1,111 @@
+(ns jepsen.control.k8s
```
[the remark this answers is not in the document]
pavlobaron: I believe they're used interchangealby, so it can be expected that someone who's looking for "kubernetes" will also be looking fof k-eight-s or "k8s". I'm open here, too.

jepsen/src/jepsen/control/k8s.clj
pavlobaron: in my test, there is some complexity behing "--". Effectively, I do believe they have deprecated the previous way of exec-ing commands and are replacing it now with the full freedom exactly behind "--" for all kinds of special cases. Can you give an example of whatever kind? it's easy to run against any k8s context, so I can quickly verify. Thanks.

jepsen/src/jepsen/control/k8s.clj
pavlobaron: Docker remote has it as def, so does ssh as well. Why would it be a function in this case?

jepsen/src/jepsen/control/k8s.clj
aphyr: Like I said, single and double quotes, semicolons, ampersands, dashes, pipes, redirection brackets, all that good stuff. You'll wanna make sure that gets correctly escaped, or this could end up exploding in some really awkward ways. Like, delete-your-home-directory kinda stuff, haha. :)

jepsen/src/jepsen/control/k8s.clj
aphyr: I missed that in the docker remote, but it shouldn't do it either. Trying to cut down on extraneous compile-time constants, and also leave ourselves escape hatches in case things need to be parameterized later.

jepsen/project.clj
```diff
@@ -1,4 +1,4 @@
-(defproject jepsen "0.2.1-SNAPSHOT"
+(defproject jepsen "0.2.2-SNAPSHOT"
```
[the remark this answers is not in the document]
pavlobaron: fixed

jepsen/src/jepsen/control/k8s.clj
pavlobaron: unfortunately, I don't see how to formalize that in local tests without a significantly larger effort, but I'll do some in repl and report the results here for comparison.

jepsen/src/jepsen/control/k8s.clj
pavlobaron: fixed

jepsen/src/jepsen/control/k8s.clj
pavlobaron: that was a debug left-over, it's gone now

jepsen/src/jepsen/control/k8s.clj
pavlobaron: `foobar=> (k8s/exec "" "" "config-56ff747f6b-rqsjz" {:cmd "export FOOBAR=tmp && cd /`echo $FOOBAR` ; ls -la $(pwd)"})
{:exit 0, :out "total 8\ndrwxrwxrwt 1 root root 4096 Jul 17 19:24 .\ndrwxr-xr-x 1 root root 4096 Jul 16 15:49 ..\n", :err ""}
foobar=>`

pavlobaron: lol, I spent more time escaping here int the inline editor than running the actual test, but example above should get the point across: here, I'm running remote's exec function with all kinds of linux shell trickery / special characters, as if it would be called from controls/sh itself, against some random pod in my currently default k8s context/namespace. I believe on the way :cmd gets properly escaped already, the example anyway demostrates correctness IMO - we're working with an env FOOBAR to hold temp-folder, in which we change and from which we ls, all as expected.

Also, double-dashing sh -c after a kubectl exec is to my knowledge the safest way to avoid weird inputs and outputs.

aphyr: Cool cool. :)

aphyr:
I think this is good to go. Thanks @pavlobaron!

PR #502, opened 2021-04-03
Fix truncate nemesis lambda arguments

alesapin:
It seems like `control/on-nodes` function must accept lambda with two arguments.


aphyr:
Good catch, thanks!

PR #508, opened 2021-07-07
Switched off privileged mode to db node containers.

### LINKED ISSUE #506: bin/up logs out on Pop!_OS
Hi,

I'm trying to run jepsen docker setup on Pop!_OS:

```
$ cat /etc/os-release
NAME="Pop!_OS"
VERSION="20.10"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 20.10"
VERSION_ID="20.10"
HOME_URL="https://pop.system76.com"
SUPPORT_URL="https://support.system76.com"
BUG_REPORT_URL="https://github.com/pop-os/pop/issues"
PRIVACY_POLICY_URL="https://system76.com/privacy"
VERSION_CODENAME=groovy
UBUNTU_CODENAME=groovy
LOGO=distributor-logo-pop-os
```
Whenever I run `bin/up` I get logged out. Containers are created and started, but as soon as all of them are started I get logged out. I'm unable to log in, so I need to restart the machine. After the next login, I'm able to start the control container using `docker container start`, but starting any of the `node` containers gets me logged out again. After the next login, I'm able to start all containers and everything works as expected.

I was able to track down that starting containers in `privileged mode` creates this issue. After tweaking docker-compose template to create docker-compose file without privileged mode (but mapping volumes with - "/sys/fs/cgroup:/sys/fs/cgroup:ro") jepsen is able to start without any logging out / restarting.
Of course, without priveleged mode, it is not possible to run nemesis, so this is not an option.

However, running `jgoerzen/debian-base-standard` in privileged mode alone using
```
docker run -td --stop-signal=SIGRTMIN+3 \
  --tmpfs /run:size=100M --tmpfs /run/lock:size=100M \
  -v /sys/fs/cgroup:/sys/fs/cgroup:ro \
  --name=name --privileged jgoerzen/debian-base-whatever
```
is not a problem.

One colleague of mine, who is using the same OS, tried running jepsen and the same happens. On Windows WSL and Mac_OS, everything works fine.

Cheers,
Milan

m1l4n54v1c:
Resolves #506

### DIFF
--- a/docker/template/docker-compose.yml
+++ b/docker/template/docker-compose.yml
@@ -5,26 +5,17 @@ x-node:
   env_file: ./secret/node.env
   secrets:
     - authorized_keys
-  privileged: true
   tty: true
   tmpfs:
     - /run:size=100M
     - /run/lock:size=100M
   volumes:
-    # This now appears to break systemd, as of 2020/01/25, for reasons I don't
-    # understand?
-    # jepsen-n9  | Failed to create /init.scope control group: Read-only file system
-    # jepsen-n9  | Failed to allocate manager object: Read-only file system
-    # - "/sys/fs/cgroup:/sys/fs/cgroup:ro"
-    #
-    # Let's try it without ro
-    # - "/sys/fs/cgroup:/sys/fs/cgroup"
-    #
-    # OK, that ALSO breaks. But we can apparently leave out the cgroup mount
-    # entirely? I wonder what horrible things happen because of that...
+    - "/sys/fs/cgroup:/sys/fs/cgroup:ro"
     - "jepsen-shared:/var/jepsen/shared"
   networks:
     - jepsen
+  cap_add:
+    - NET_ADMIN
   ports:
     - ${JEPSEN_PORT:-22}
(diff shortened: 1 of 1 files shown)


aphyr:
Wooo, let's hope this works!

aphyr:
So this breaks the docker enviroment on my Debian machine (5.9.0-5-amd64 #1 SMP Debian 5.9.15-1 (2020-12-17) x86_64 GNU/Linux, Docker 20.10.10)--every node gets stuck complaining:

```
jepsen-n5  | Failed to create /init.scope control group: Read-only file system
jepsen-n5  | Failed to allocate manager object: Read-only file system
jepsen-n5  | [!!!!!!] Failed to allocate manager objec
```

And then crashes. Any idea how to fix this?

m1l4n54v1c:
Hmmm, it works on my machine :). 

I guess it could be related to [systemd "issue"](https://github.com/systemd/systemd/issues/19245). I'm running `systemd 247`. What's your version?

PR #509, opened 2021-07-16
Adding ALL capabilities to node containers

m1l4n54v1c:
NETWORK_ADMIN capability is not sufficient for all nemesis. For example, clock_scrambler needs the capability to mess with the date. Adding ALL capabilities to node containers.


aphyr:
Thanks!

PR #510, opened 2021-07-19
translate the jepsen doc with chinese language

jaydenwen123:
A few months ago, our team started developing a distributed system component, which is based on raft algorithm. Then we need to find a way to verify the correctness of our system. Fortunately, we find the jepsen, a very excellent component to achieve this goal. Unluckily, there is little Chinese reference about jepsen. Therefore, we have taken much time to learn jepsen document and run jepsen to test our system.

After finishing our work, we thought we could do some work to help those who are facing the same problem. So we decided to translate the jepsen document into Chinese. Now the work has been finished and saved in the dir: jepsen/docs/cn_tutorial/.


aphyr:
Hi there! I'm afraid I have no way to maintain or support this, but you're welcome to maintain it independently. :-)

PR #518, opened 2021-12-02
Fix docker scripts

zouyonghao:
1. Fix dom-top dependency
    It seems there is no `dom-top 1.0.7-SNAPSHOT` on clojars https://repo.clojars.org/dom-top/dom-top/, so we build it ourselves.

2. fix ignored source files in jepsen/store
    The `*/store/**` files are ignored in `.gitignore`, causing some source files in `jepsen/src/jepsen/store/` are ignored too.

### DIFF
--- a/jepsen/src/jepsen/web.clj
+++ b/jepsen/src/jepsen/web.clj
@@ -65,7 +65,6 @@
   (let [cached @test-cache
         ; Drop the most recent n keys from the cache--these we assume may have
         ; changed recently.
-        _ (info :cached cached)
         cached (->> (keys cached)
                     (take test-cache-mutable-window)
                     (reduce dissoc cached))]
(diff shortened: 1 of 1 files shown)


aphyr:
Can you split these up, please? The gitignore issue seems real (I'm guessing tar mis-interprets the * as operating at any depth, not just the top level), but the dom-top issue is easy to fix; I just need to cut a snapshot release for Clojars.

zouyonghao:
> Can you split these up, please? The gitignore issue seems real (I'm guessing tar mis-interprets the * as operating at any depth, not just the top level), but the dom-top issue is easy to fix; I just need to cut a snapshot release for Clojars.

Sure, see https://github.com/jepsen-io/jepsen/pull/519

aphyr:
Thanks! You'll find a fresh release of dom-top on clojars now too. :)

PR #522, opened 2021-12-07
update zookeeper to latest jepsen

zouyonghao:


zookeeper/src/jepsen/zookeeper.clj
```diff
@@ -112,21 +118,23 @@
...
+                                   {:type :info, :f :stop}]))
                           (gen/time-limit 15))
           :model   (model/cas-register 0)
           :checker (checker/compose
-                     {:perf   (checker/perf)
-                      :linear checker/linearizable})}))
+                    {:perf   (checker/perf)
+                     :linear (checker/linearizable 
+                              {:model (model/cas-register)
+                               :algorithm :linear})})}))
```
aphyr, line 137: If you haven't benchmarked, might be best to let knossos pick the competitive algo here. Or port the workload to Elle, which should be *dramatically* faster.

aphyr:
This looks good! Hey, if you're poking at this--we *should* expect Zookeeper to actually violate linearizability sometimes, so if anomalies crop up in your experiments, that's fantastic. :-)

zouyonghao:
> This looks good! Hey, if you're poking at this--we _should_ expect Zookeeper to actually violate linearizability sometimes, so if anomalies crop up in your experiments, that's fantastic. :-)

Yes, I got some anomalies that I want in my experiments.

PR #525, opened 2021-12-09
[WIP] Stabilise failing tests

qvad:
According to http://jepsen-io.github.io/jepsen/jepsen.generator.html
`stagger, delay, etc. now take total rates, rather than the rate per thread.`

All values changes approximately from 1/x to 1/(x* n), where n is number of threads


qvad:
sorry, miss click

PR #526, opened 2022-01-20
Update docker scripts for newer Docker/Debian

dancmeyers:
- On Mac `/bin/echo -e "<control chars>"` prints out `-e <uninterpreted
  control chars>` to the terminal. This seems to happen in the latest
  Debian image too. Use `printf` for compatibility across both and
  proper colouration.
- Check that `POSITIONAL` has some elements before trying to `set` with
  it, to avoid `unbound variable` error with `set -o nounset`.
- Remove cgroup sharing from host, and make container privileged, to
  allow nodes to spin up with cgroupv2. Docker > 20.10 will
  automatically create a private cgroup slice in this case.

### DIFF
--- a/docker/bin/up
+++ b/docker/bin/up
@@ -9,15 +9,15 @@ set -o nounset
 # set -o xtrace
 
 ERROR() {
-    /bin/echo -e "\e[101m\e[97m[ERROR]\e[49m\e[39m" "$@"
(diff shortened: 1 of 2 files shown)


aphyr:
Thanks @dancmeyers! :-)

PR #538, opened 2022-06-06
WIP: Web: fix datetime parsing so fresh test runs correctly appear

24twelve:
Hi.

I live in UTC+5:00 timezone. I've been learning to use Jepsen, when I encountered bug: my fresh test runs would not appear in Jepsen UI.

Screenshot_846

I've tried to fix this. The idea is to always operate UTC datetime inside the application and convert to local time only for user output. 
I've built it locally and it works for me:

Screenshot_845

Could you have a look on this PR? I am totally not familiar with clojure and any changes would be welcome.

### DIFF
--- a/jepsen/src/jepsen/web.clj
+++ b/jepsen/src/jepsen/web.clj
@@ -70,7 +70,7 @@
 (defn parse-time
   "Parses a time from a string"
   [t]
-  (-> (timef/parse-local basic-date-time t)
+  (-> (timef/parse basic-date-time t)
       time.coerce/to-date-time))
 
 (defn fast-tests
@@ -150,7 +150,7 @@
   (let [r    (:results t)
         time (->> t
                   :start-time
-                  (timef/unparse (timef/formatters :date-hour-minute-second)))]
+                  (timef/unparse-local (timef/formatters :date-hour-minute-second)))]
     [:tr
      [:td [:a {:href (url t "")} (:name t)]]
      [:td [:a {:href (url t "")} time]]


aphyr:
Ha! That's neat! The timezone math in Jepsen is a total hack, largely there by accident, and I've been meaning to do something just like this. On the other hand, I'm a little worried by the fact that the rendered times in the web view here seem off by an hour... Does that seem right to you?

aphyr:
I'm also sort of confused by the fact that some test runs aren't showing at *all*--why are we getting any kind of truncation behavior in the first place? Surely they're parsable...

24twelve:
`why are we getting any kind of truncation behavior in the first place`
That is the point. If I run tests, I cannot see them in web afterwards. This is what I tried to fix.
I believe that the problem is somewhere [here](https://github.com/jepsen-io/jepsen/blob/main/jepsen/src/jepsen/web.clj#L181). But currently I got stuck.

I will return to this issue later if I find time.

aphyr:
Ah, no, that'd do it! I think what we really need there is a far-future time. Infinite would be ideal, but an easy hack would be to add, I dunno, a thousand years to the present. ;-)

24twelve:
Actually, proposed solution does break something: if I click on any link in test row, web returns `404 not found`. Please, do not merge until I resolve this :)

24twelve:
After some digging, I realized, that problem is fixed and does not repoduce in Jepsen v0.2.7.
Sorry for stealing your attention.

PR #540, opened 2022-07-02
fix `view source` link in generated API docs (codox source-uri)

nurturenature:
`view source` links in generated API docs are missing **v**{version}:

- 404 without the **v**: https://github.com/jepsen-io/jepsen/blob/0.2.7/jepsen/src/jepsen/checker.clj#L737

- work with the **v**:  https://github.com/jepsen-io/jepsen/blob/v0.2.7/jepsen/src/jepsen/checker.clj#L737


aphyr:
Thanks!

PR #550, opened 2022-08-23
packet-nemesis: use tc filters to only shape to/from traffic of `:targets` db nodes 

nurturenature:
The original packet-nemesis is a bit simplistic.
It adds a netem qdisc to `:targets` db nodes that affect all outbound traffic.
This includes the control node and ignores any effects on inbound traffic from other nodes.

In the original PR merge discussion
you provided a [sample gist](https://gist.github.com/aphyr/6d40e0febf25d12653b0abb868b6bf9e) showing how to have a standard priority queue for good traffic,
and use filters to a netem queue for bad traffic. 

Switching to those techniques in this PR, the packet-nemesis will effect the inbound traffic to, and the outbound traffic from `:targets` db nodes only.

Control node traffic remains reliable.

Affecting both inbound traffic to and outbound traffic from `:targets` seems more natural.
Can use `:all` for whole network.
More aligned with how other packages work, these `:targets` will be impacted, have this relationship with the other nodes, etc.

The log messages are more like other packages:
```clj
:start-packet   [:one {:delay {}, :rate {:rate :1mbit}}]
:start-packet   [:shaped {"n1" #{"n5"}, "n2" #{"n5"}, "n3" #{"n5"}, "n4" #{"n5"}, "n5" #{"n2" "n1" "n4" "n3"}} :netem [:delay :50ms :10ms :25% :distribution :normal :rate :1mbit]]
...
:stop-packet    nil
:stop-packet    [:reliable {"n1" nil, "n2" nil, "n3" nil, "n4" nil, "n5" nil}]
```

To use random property style with packet-package:
```clj
{:packet {:targets (:db-targets opts)
          :behaviors (for [_ (range 10)]
                       (->> net/all-packet-behaviors
                            (random-sample (/ (+ 1 (rand-int 3))
                                              (count net/all-packet-behaviors)))
                            (into {})))}}
```

More comprehensive, test-all style with packet-nemesis:
```clj
; Every combination of network latency and rate limit
; effecting the entire network:
(->> (for [rate    (:net-rates test)
           latency (:net-latencies test)]
       ...
       {:type  :info,
        :f     :start-packet,
        :value [:all {:delay {:time latency}
                      :rate  {:rate rate}}]}))
```

Limping hardware/node:
```clj
; "n1" progressively getting worse...
(gen/flip-flop (map (fn [delay-time corrupt]
                      {:type  :info
                       :f     :start-packet
                       :value [["n1"]
                               {:delay   {:time    (str delay-time "ms")}
                                :corrupt {:percent (str corrupt "%")}}]})
                    (cycle (range 100 1001 100))
                    (cycle (range 10 101 10)))
                (repeat {:type  :info
                         :f     :stop-packet
                         :value nil}))
```

Keep pushing a majority further and further apart until consensus isn't practical/possible.
How far apart is that how does cluster fail?:
```clj
(map (fn [delay]
       {:type :info,
        :f    :start-packet,
        :value [:majority {:delay {:time delay}}]})
     (range))
```

At minimum, I would like to fix the issue of control node traffic being impacted.
I also think using filters for both to/from `:targets` nodes is more package'y and a slightly better fit for common scenarios.

Is this Ok?
(I think it's closer to what I should have originally submitted a couple of weeks ago.)

Or any desired changes?

Thanks!

### DIFF
--- a/jepsen/src/jepsen/net.clj
+++ b/jepsen/src/jepsen/net.clj
@@ -65,7 +65,7 @@
   [dev]
   (try+
    (su (exec tc :qdisc :del :dev dev :root))
-   (catch [:exit 2] {}
+   (catch [:exit 2] _
      ; no qdisc to del
      nil)))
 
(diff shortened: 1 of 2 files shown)


nurturenature:
P.S. A plot showing n1 in a data center with a failing network. The generator swings the counter back and forth. n1's counter gets quite stale, ending up invalid and also influencing n5.

```clj
:final-reads
   (["n1" -188682]
    ["n2" -172371]
    ["n3" -172371]
    ["n4" -172371]
    ["n5" -200711]),
   :counter [[-172371 -172371]]
```

image

aphyr:
> Affecting both inbound traffic to and outbound traffic from :targets seems more natural.

One thing you might want to consider is what happens when a user requests a simple case like "add 10 ms of latency to every node"--they'll actually get 20 ms of latency between nodes, right? Would it make sense to just apply this to inbound packets, the same as the network partition nemesis?

aphyr:
Oh, yeah, I see now. The flip case is pulling a single node far away, and because the targeter can affect just one node, it makes sense to be symmetric there. I guess conceptually we can think of this not as a directly-connected topology, but like "lengthening the cable" to each node. That makes it feel more natural that you'd get twice the latency when the targets happen to be multiple nodes.

jepsen/src/jepsen/net.clj
```diff
@@ -46,75 +48,118 @@
 
 (def tc "/sbin/tc")
 
+(defn net-dev
```
aphyr, line 51: OH hey this is great--I was *just* hitting issues with some other Jepsen code that assumed "eth0" as the primary iface. Thanks for this!

jepsen/src/jepsen/net.clj
```diff
@@ -46,75 +48,118 @@
...
+    (assert iface
+            (str "Couldn't determine network interface!\n" choices))
+    iface))
+
+(defn qdisc-del
+  "Deletes root qdisc for given dev on current node."
+  [dev]
+  (try+
+   (su (exec tc :qdisc :del :dev dev :root))
+   (catch [:exit 2] {}
```
aphyr, line 68: Is this {} meant to be _? Usually it'd be a binding form there, right?

jepsen/src/jepsen/net.clj
```diff
@@ -46,75 +48,118 @@
...
+               (concat args [behavior percent correlation])
+               :rate
+               (concat args [:rate rate])))
+           [])))
+
+(defn- net-shape!
+  "Shared convenience call for iptables/ipfilter.
+   Shape the network with tc qdisc, netem, and filter(s) so target nodes have given behavior."
+  [_net {:keys [nodes] :as _test} targets behavior]
+  (let [results (on-many nodes
```
aphyr, line 125: BTW, using `(c/on-nodes test nodes (fn [test node] ...))` is a bit faster, since it'll re-use the existing SSH sessions for the test.

aphyr:
Great work, @nurturenature. I've got a few minor comments here, but nothing serious. Happy to merge as soon as you're ready. :-)

jepsen/src/jepsen/net.clj
nurturenature: The function is from the net-flaky-nemesis.clj gist you provided. :)

jepsen/src/jepsen/net.clj
nurturenature: yes, good catch.

jepsen/src/jepsen/net.clj
nurturenature: Noticeably faster, thanks!

nurturenature:
Updated to (on-nodes) vs (on-many) and fixed the binding form syntax.

Think everything is ready.

Thanks so much!

aphyr:
Great, here we go! Thanks again @nurturenature :-)

jepsen/src/jepsen/net.clj
aphyr: Ha! That tells you something about how flaky my memory is, lol ;-)

PR #551, opened 2022-08-31
add nemesis.combined/file-corruption-package, enhance nemesis/(bitflip, truncate-file)

nurturenature:
This PR proposes to add file corruption to `nemesis/combined`,
and in doing so enhance `bitflip` and `truncate-file` nemeses.

----

Hopefully the docs are a clear explanation:

### file-corruption-package

Opts:

```clj
{:file-corruption
 {:targets     [...] ; A collection of node specs, e.g. [:one, [\"n1\", \"n2\"], :all]
  :corruptions [     ; A collection of file corruptions, e.g.:
   {:bitflip "/path/to/file"
    :probability 1e-3},
   {:bitflip "/path/to/dir"
    :probability [1e-3 1e-4 1e-5]},
   {:truncate "/path/to/file/or/dir"
    :drop (range 1 1025)}]}}
```
If a directory is given to `:bitflip` or `:truncate`,
a new random file is selected from that directory on each target node for each operation.

If `:probability` or `:drop` is a collection, a new `rand-nth` value is chosen for each operation.

----

In the logs:
```clj
:nemesis :info :bitflip {"n1" {:file "/root/antidote/data_riak_core", :probability 1.0E-4}}
:nemesis :info :bitflip {"n1" {:file "/root/antidote/data_riak_core/riak_core_ring.default.20220831033635", :probability 1.0E-4}}
...
:nemesis :info :truncate {"n1" {:file "/root/antidote/data_antidote", :drop 269}, "n2" {:file "/root/antidote/data_antidote", :drop 269}}
:nemesis :info :truncate {"n1" {:file "/root/antidote/data_antidote/1027618338748291114361965898003636498195577569280--1027618338748291114361965898003636498195577569280.LOG", :drop 269}, "n2" {:file "/root/antidote/data_antidote/365375409332725729550921208179070754913983135744--365375409332725729550921208179070754913983135744.LOG", :drop 269}}
```
----

The generator functions need to be lightweight.

The existing `bitflip` and `truncate-file` take a specific `:file` to operate on. This forces any random selection of files into the generator.
Doing operations like `file?` and `ls-full` from a generator are way too heavyweight side-effects. Jepsen gets *quite* weird when generators take a lot of time and interact with other parts of the system.

Enhancing `:file` to be either a file or a directory to select a random file from allows the generator to stay lite and the nemesis to do the needed work.
It's also backwards compatible.

----

### bitflip

A nemesis which introduces random bitflips in files. Takes operations like:
```clj
{:f     :bitflip
 :value {"some-node" {:file        "/path/to/file or /path/to/dir"
                      :probability 1e-3}}}
```
This flips 1 x 10^-3 of the bits in `/path/to/file`, or a random file in `/path/to/dir`, on "some-node".

----

### truncate-file

A nemesis which responds to
  ```clj
  {:f     :truncate
   :value {"some-node" {:file "/path/to/file or /path/to/dir"
                        :drop 64}}}
  ```
  where the value is a map of nodes to `{:file, :drop}` maps, on those nodes,
  drops the last `:drop` bytes from the given file, or a random file from the given directory.

----

One other misc change was to (re)enable `cu/file?`.
It was deprecated 6 years ago for `cu/exists?`.
If that's not Ok it can be moved inline.

----

My initial usage:

```clj
:file-corruption {:targets     [["n1"]]
                  :corruptions [{:bitflip db/node-antidote-data-dir
                                 :probability [1e-2 1e-3 1e-4]}
                                {:bitflip db/node-antidote-riak-dir
                                 :probability [1e-2 1e-3 1e-4]}]}
```

With the random generators finding a tender target:

```clj
:nemesis :info :bitflip {"n1" {:file "/root/antidote/data_antidote", :probability 1.0E-4}}
:nemesis :info :bitflip {"n1" {:file "/root/antidote/data_antidote/a_stable_meta_data_table", :probability 1.0E-4}}
```

To replicate what the etcd test does for file corruption:

```clj
:file-corruption {:targets     [["n1" "n2"]] ; or a minority function
                  :corruptions [{:bitflip "/member/wal"
                                 :probability [1e-3 1e-4 1e-5]}
                                {:bitflip "/member/snap"
                                 :probability [1e-3 1e-4 1e-5]}
                                {:truncate "/member/wal"
                                 :drop (range 1 1025)}]}
```

I was pretty sure that a `file-corruption-package` would be welcome so went ahead with a proposed implementation that tried to fit into the existing framework and usage. Am open to redoing anything that doesn't fit.

Is this too much? A good idea? Should be done differently?

Thanks!

### DIFF
--- a/jepsen/src/jepsen/control/util.clj
+++ b/jepsen/src/jepsen/control/util.clj
@@ -30,10 +30,13 @@
             opts))))
 
 (defn file?
+  "Is `filename` a regular file that exists?"
   [filename]
-  (throw (RuntimeException. "Use exists? instead; file? will be used to tell if
-                            something is a file, as opposed to a directory or
-                            link.")))
+  (try+
+   (exec :test :-f filename)
+   true
+   (catch [:exit 1] _
+     false)))
 
 (defn exists?
   "Is a path present?"
--- a/jepsen/src/jepsen/nemesis/combined.clj
+++ b/jepsen/src/jepsen/nemesis/combined.clj
@@ -22,6 +22,7 @@
                     [net :as net]
                     [util :as util :refer [majority
                                            minority-third
+                                           rand-distribution
                                            random-nonempty-subset]]]
             [jepsen.nemesis.time :as nt]))
 
@@ -367,18 +368,24 @@
    {:file-corruption
     {:targets     [...] ; A collection of node specs, e.g. [:one, [\"n1\", \"n2\"], :all]
      :corruptions [     ; A collection of file corruptions, e.g.:
-      {:bitflip \"/path/to/file\"
+      {:type :bitflip
+       :file \"/path/to/file\"
        :probability 1e-3},
-      {:bitflip \"path/to/dir\"
-       :probability [1e-3 1e-4 1e-5]},
-      {:truncate \"path/to/file/or/dir\"
-       :drop (range 1 1025)}]}}
+      {:type :bitflip
+       :file \"path/to/dir\"
+       :probability {:distribution :one-of :values [1e-3 1e-4 1e-5]}},
+      {:type :truncate
+       :file \"path/to/file/or/dir\"
+       :drop {:distribution :geometric :p 1e-3}}]}}
    ```
    
-   If a directory is given to `:bitflip` or `:truncate`,
+   `:type` can be `:bitflip` or `:truncate`.
+
+   If `:file` is a directory,
    a new random file is selected from that directory on each target node for each operation.
 
-   If `:probability` or `:drop` is a collection, a new `rand-nth` value is chosen for each operation.
+   `:probability` or `:drop` can be specified as a single value or a `distribution-map`.
+   Use a `distribution-map` to generate a new random value for each operation using [[jepsen.util/rand-distribution]].
    
    See [[jepsen.nemesis/bitflip]] and [[jepsen.nemesis/truncate-file]].
 
@@ -388,17 +395,20 @@
         targets     (:targets     file-corruption (node-specs db))
         corruptions (:corruptions file-corruption)
         gen (->> (fn gen [_test _context]
-                   (let [targets     (->> (rand-nth targets)
-                                          (db-nodes {:nodes nodes} db))
-                         corruption  (rand-nth corruptions)
-                         {:keys [bitflip truncate probability drop]} corruption
-                         probability (->> probability (util/coll) (rand-nth))
-                         drop        (->> drop        (util/coll) (rand-nth))
-                         [f corruption] (cond
-                                          bitflip  [:bitflip  {:file bitflip  :probability probability}]
-                                          truncate [:truncate {:file truncate :drop        drop}])]
+                   (let [targets (->> (rand-nth targets)
+                                      (db-nodes {:nodes nodes} db))
+                         {:keys [type file probability drop]} (rand-nth corruptions)
+                         corruption (case type
+                                      :bitflip  (let [probability (cond
+                                                                    (number? probability) probability
+                                                                    (map? probability) (rand-distribution probability))]
+                                                  {:file file :probability probability})
+                                      :truncate (let [drop (cond
+                                                             (number? drop) drop
+                                                             (map? drop) (rand-distribution drop))]
+                                                  {:file file :drop drop}))]
                      {:type  :info
-                      :f     f
+                      :f     type
                       :value (->> targets
                                   (reduce (fn [plan node]
                                             (assoc plan node corruption))
--- a/jepsen/src/jepsen/nemesis/combined.clj
+++ b/jepsen/src/jepsen/nemesis/combined.clj
@@ -529,6 +529,6 @@
   File corruption options:
     
     :targets     A collection of node specs, e.g. [:one, :all]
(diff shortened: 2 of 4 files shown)


jepsen/src/jepsen/control/util.clj
```diff
@@ -30,10 +30,13 @@
...
+  "Is `filename` a regular file that exists?"
   [filename]
-  (throw (RuntimeException. "Use exists? instead; file? will be used to tell if
-                            something is a file, as opposed to a directory or
-                            link.")))
+  (try+
+   (exec :test :-f filename)
+   true
+   (catch [:exit 1] _
+     false)))
```
aphyr, line 39: Yup, I'm all for this. Been more than long enough. :-)

jepsen/src/jepsen/nemesis/combined.clj
```diff
@@ -359,6 +359,60 @@
...
+   ```clj
+   {:file-corruption
+    {:targets     [...] ; A collection of node specs, e.g. [:one, [\"n1\", \"n2\"], :all]
+     :corruptions [     ; A collection of file corruptions, e.g.:
+      {:bitflip \"/path/to/file\"
+       :probability 1e-3},
+      {:bitflip \"path/to/dir\"
+       :probability [1e-3 1e-4 1e-5]},
+      {:truncate \"path/to/file/or/dir\"
+       :drop (range 1 1025)}]}}
```
aphyr, line 375: This `(range)` concerns me a bit, because it's easy to imagine someone actually materializing and iterating over, say, (range 1 1e10), which... could be disastrous, especially if it leaks into a pretty-printer somewhere. You also might not want a uniform distribution over the choices--you might want to lose just one byte sometimes, and hundreds of megs other times, and a uniform distribution would almost never select small truncations. Maybe something like

```clj
{:distribution :geometric
 :mean 500}
```

or 

```clj
{:distribution :uniform
 :min 50
 :max 5000}
```

jepsen/src/jepsen/nemesis/combined.clj
```diff
@@ -359,6 +359,60 @@
...
+  [{:keys [faults db file-corruption interval nodes] :as _opts}]
+  (let [needed?     (:file-corruption faults)
+        targets     (:targets     file-corruption (node-specs db))
+        corruptions (:corruptions file-corruption)
+        gen (->> (fn gen [_test _context]
+                   (let [targets     (->> (rand-nth targets)
+                                          (db-nodes {:nodes nodes} db))
+                         corruption  (rand-nth corruptions)
+                         {:keys [bitflip truncate probability drop]} corruption
+                         probability (->> probability (util/coll) (rand-nth))
```
aphyr, line 395: Yeah this rand-nth is gonna do linear traversal over the ranges. Honestly we do enough probabilistic selection stuff that it might be worth having a standard function in jepsen.util for drawing a random value given a map like {:distribution :uniform, :min 1, :max 50}. Then we can just say that probability and drop take distribution maps.

aphyr (commented):
I think this is on the right track! I'd like to change a bit about how the parameters are named/structured, but the general logic is sound!

jepsen/src/jepsen/control/util.clj
nurturenature: Oh good, as I've been using it in other places too. :-)

jepsen/src/jepsen/nemesis/combined.clj
```diff
@@ -359,6 +359,60 @@
...
 
+(defn file-corruption-package
+  "A nemesis and generator package that corrupts files.
+   
+   Opts:
+   ```clj
+   {:file-corruption
+    {:targets     [...] ; A collection of node specs, e.g. [:one, [\"n1\", \"n2\"], :all]
+     :corruptions [     ; A collection of file corruptions, e.g.:
+      {:bitflip \"/path/to/file\"
```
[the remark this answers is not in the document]
nurturenature: Yes, it's more explicit and self documenting. Also a bit better if new `:type`s are added.

Will commit and summarize in main conversation thread.

jepsen/src/jepsen/nemesis/combined.clj
nurturenature: Yes, it's  problematic Clojure and only allows for a uniform distribution.

Will commit and summarize in main conversation thread.

jepsen/src/jepsen/nemesis/combined.clj
nurturenature: > might be worth having a standard function in jepsen.util for drawing a random value given a map like {:distribution :uniform, :min 1, :max 50}

That's very expressive.
Will encourage use of more than a uniform distribution too.

Will commit and summarize in the main conversation thread.

nurturenature:
Thanks for such a complete and helpful review!

And as a not stats savvy enough person myself, I think creating `(util/rand-distribution)` may help encourage more appropriate choices for random values in configs.

aphyr:
Two weeks! Gosh I am SO sorry. It's not you, it's me!

Yes, I think this is great. I paused a bit on `rand-distribution`, because it's returning a sample *from* a distribution, not a distribution itself, but on second thought it fits well with other kinds of rand: we're saying this generates a rand, and distribution is the mechanism. And yes, I agree, there are lots of things in Jepsen that need this, and it'll be very helpful to have!

PR #553, opened 2022-09-06
docker: Update debian & jdk versions

Fumesover:
Debian buster will reach its EOL soon (source: https://wiki.debian.org/DebianReleases), bumping the dependency to the last release


aphyr:
Thanks Fumesover!

PR #554, opened 2022-09-28
Use `strict-host-key-checking` argument in SSHJ and SCP

antonio2368:
`strict-host-key-checking`argument is ignored in SSHJ as far as I see.
It should be used correctly now both in SSHJ and SCP.


aphyr:
Ah, wonderful! Thank you. :-)

PR #567, opened 2023-03-06
jepsen.net/net-dev: (more) correctly identify the network interface by using ip vs /sys/class/net

nurturenature:
Hi,

`jepsen.net/net-dev` currently identifies the network interface by looking in `/sys/class/net`.
Previous to that it was hard coded to `eth0`.

This works well with LXC containers and/or VMs.

In other container environments, depending on their config and how they're started, looking in  `/sys/class/net` can be misleading.
It can show `docker0`, `veth-this`, `veth-that`, etc., in addition to the actual container interface.

`ip` is network namespace aware and will only show interfaces that can actually be used in the container namespace.

This has been tested with Debian/LXC and Ubuntu in Docker with namespaces.

----

And then a misc found during debugging:

`jepsen.control.util/wget-helper!` needs to `throw+` on error as the exec result is a map, not an Exception.

This allows a wget failure to present as an Exception with the result map for debugging, vs a class cast Exception.

----

And if this PR is accepted, could it also be followed with a `0.3.2-SNAPSHOT` to clojars?

Thanks!

### DIFF
--- a/jepsen/src/jepsen/control/util.clj
+++ b/jepsen/src/jepsen/control/util.clj
@@ -124,7 +124,7 @@
                 (catch [:type :jepsen.control/nonzero-exit, :exit 4] e
                   (if (pos? tries)
                     ::retry
-                    (throw e))))]
+                    (throw+ e))))]
       (if (= ::retry res)
         (recur (dec tries))
         res))))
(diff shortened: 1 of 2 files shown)


aphyr:
Yeah, this is very sensible, thank you! Portability always appreciated. :-)

aphyr:
Cut a 0.3.2-SNAPSHOT release for you. Pardon my delay, as usual. I have a good excuse this time! Was moving and my computer was in boxes! ;-)

nurturenature:
> Cut a 0.3.2-SNAPSHOT release for you.

Thanks!

I will update the CI to use it on Monday.

PR #569, opened 2023-03-21
Fix clock changes in latest Jepsen

### LINKED ISSUE #568: Docker clock changes cause test failures
We (MongoDB) have seen behavior in recent versions of Jepsen that cause test failures in about 60% of runs due to odd clock skew. Specifically, the clock is incorrect between runs, which causes subsequent `apt-update` commands in the containers to fail due to certificate validation failure. Our Jepsen tests run with clock skew disabled, so we aren't sure *why* the clock is being changed, but it looks like the recent change to make test node containers `privileged` and have `ALL` capabilities is *how* it's happening.

Interestingly, this only seems to be a problem on test hosts that have an NTP client running. When we run the tests on our virtual workstations, which do not have an NTP client running, the tests succeed. It seems that the clock skew in the test node containers is racing with the NTP client somehow, which causes the observed failures. However, as stated above, we have not been able to determine so far why the clock skew occurs in the first place.

deriamis:
This partially reverts a previous commit, which added `privileged: true` and `capabilities: ALL` to Docker test nodes. It also adds mounts which should ensure the containers come up with the same local time as the host.

Closes #568

### DIFF
--- a/docker/template/docker-compose.yml
+++ b/docker/template/docker-compose.yml
@@ -10,12 +10,14 @@ x-node:
     - /run:size=100M
     - /run/lock:size=100M
   volumes:
+    - "/sys/fs/cgroup:/sys/fs/cgroup:ro"
+    - "/etc/timezone:/etc/timezone:ro"
+    - "/etc/localtime:/etc/localtime:ro"
     - "jepsen-shared:/var/jepsen/shared"
   networks:
     - jepsen
-  privileged: true
   cap_add:
-    - ALL
+    - NET_ADMIN
   ports:
     - ${JEPSEN_PORT:-22}
 
@@ -44,5 +46,7 @@ services:
     networks:
       - jepsen
     volumes:
+      - "/etc/timezone:/etc/timezone:ro"
+      - "/etc/localtime:/etc/localtime:ro"
       - "jepsen-shared:/var/jepsen/shared"
 %%DBS%%


aphyr:
Hmm. @dancmeyers, since you added some of these parameters specifically to fix earlier broken-ness in Docker, do you want to weigh in here? I don't want to flip back and forth between two broken states if we can avoid it.

deriamis:
@aphyr By the way, this PR is just what we applied locally to prevent clock skew in the test nodes from propagating to the host. We had the clock skew nemesis disabled in our tests, so we don't know why it was happening. If someone can figure that out and prevent it, I think it would be a better fix than this.

nurturenature:
The current flipped and flopped values for:
```yaml
volumes:
  # no /sys/fs/cgroup
privileged: true
cap_add:
  - ALL
```

Evolved to that state:
- trying to workaround
  - a docker-compose regression, docker/compose#8167
  - docker's (mis)handling of cgroups, systemd 
- keep working with macOS

The docker-compose regression has been fixed but not yet released by docker.

In general, Jepsen should be able to follow [docker-debian-base](https://salsa.debian.org/jgoerzen/docker-debian-base#container-invocation-systemd-containers-busterbullseyesid):
```bash
# For a host running bullseye, or a newer cgroups and systemd, you have to use this:

docker run -td --stop-signal=SIGRTMIN+3 \
  --tmpfs /run:size=100M --tmpfs /run/lock:size=100M \
  -v /sys/fs/cgroup:/sys/fs/cgroup:rw --cgroupns=host \
  --name=name jgoerzen/debian-base-whatever
```

In the past, the proposed changes to:
```yaml
volumes:
  - "/sys/fs/cgroup:/sys/fs/cgroup:ro"
cap_add:
  - NET_ADMIN
```
Have broken some environments.

And:
> I don't want to flip back and forth between two broken states if we can avoid it

Seems to have been true for some time.

----

As for the container clock skew you are seeing, what a good reminder that clock skew is real and can be consequential!

Would it be too simplistic to defensively?
```clj
(jepsen.os/setup!)

; Try to stop ntpd service in case it is present and running.
(try
  (c/su
    (c/exec :service :ntpd :stop))
  (catch RuntimeException e))
```

dancmeyers:
I have to admit it's been so long since I touched this that I have nothing useful to add right now. Although we do have some more Jepsen work upcoming, so hopefully it doesn't bite me again then ;)

aphyr:
So--it sounds like I should close this PR then, to avoid re-breaking Docker in other environments? This may be one of those things where a fork is preferable...

nurturenature:
At the current time, I don't think it's possible to have a `compose.yml` that is cross os, recent docker versions, common hosting configs, for os style containers like Jepsen uses that works for all.

So, would recommend being conservative for now, leaving `/sys/fs/cgroup, privileged, cap_add` and then adapting to new docker versions as they are released.

nurturenature:
Docker released new versions earlier this week that better support OS/systemd containers like Jepsen uses for nodes.

It's now possible to use unprivileged containers.
See Jepsen PR #570.
It should address this issue too.

P.S. :+1:  for a CI with Jepsen.

PR #570, opened 2023-04-06
docker compose has fixed its cgroupns host configuration bug(s), improved support for OS/systemd containers

nurturenature:
I know, docker, but this time a bit of better news.

Docker released new versions earlier this week with improved support for OS/systemd containers like Jepsen uses, e.g. `jgoerzen/debian-base-standard`.

Docker compose not configuring cgroups correctly, Debian/systemd moving to v2, others moving at different paces, explains many of the recent config workarounds.

It is now possible to configure docker as recommended by [debian-base-standard](https://salsa.debian.org/jgoerzen/docker-debian-base#container-invocation-systemd-containers-busterbullseyesid) and [systemd](https://systemd.io/CONTAINER_INTERFACE/
):
```yaml
cgroup: host
volumes:
  - "/sys/fs/cgroup:/sys/fs/cgroup:rw"
# no privileged: true

# leave CAP as is to be conservative, per systemd:
#   "Do not drop CAP_SYS_ADMIN from the container"
cap_add:
  - ALL
```

With this configuration on Debian/bullseye, Jepsen works with Docker:

image

As Docker updated all of the architectures, this may be an end to (some of) the flip-flopping. I do think the config in `main` should reflect Debian stable.

----

Changes:

Node `Dockerfile` updated to `bullseye`:
```dockerfile
FROM jgoerzen/debian-base-standard:bullseye
```

`docker-compose` as an independent entity is going away and will be a plugin, `docker compose`
> [From the end of June 2023 Compose V1 won’t be supported anymore ...](https://docs.docker.com/compose/)
- change `docker-compose` -> `docker compose` 

Misc yaml linting:
- well behaved with systemd
  ```dockerfile
  ENV container=docker
  STOPSIGNAL SIGRTMIN+3
  ```
- update comments
- remove
  - `version:`
  - `MAINTAINER`

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -11,7 +11,7 @@ generate computers is *cloud native*.
 
 ## Quickstart
 
-Assuming you have docker-compose set up already, run:
+Assuming you have `docker compose` set up already, run:
 
 ```
 bin/up
--- a/docker/bin/up
+++ b/docker/bin/up
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# "To provide additional docker-compose args, set the COMPOSE var. Ex:
+# "To provide additional docker compose args, set the COMPOSE var. Ex:
 # COMPOSE="-f FILE_PATH_HERE"
 
 set -o errexit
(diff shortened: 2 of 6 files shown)


aphyr:
Woo!

PR #571, opened 2023-04-14
tune docker compose, Dockerfiles

nurturenature:
In docker compose, the control node `depends_on: n1...` so the db nodes are available for the ssh-keyscan.

I've noticed that docker compose will sometimes race the startup of the control node so the ssh-keyscan fails for some node(s). docker compose documentation:
```
Compose implementations MUST guarantee dependency services have been started before starting a dependent service.
Compose implementations MAY wait for dependency services to be “ready” before starting a dependent service.
```

So a simple `healthcheck:`, is sshd active and running?, was added to the docker-compose.yml for the db nodes.

----

docker compose was warning that it wants `jepsen-shared` to be a compose volume vs an `external:` volume due to:

```bash
# We need a fresh share volume each time we start, so we have a correct set of
# DB hosts. Why does Docker make sharing state SO hard
docker run --rm -v jepsen_jepsen-shared:/data/ debian:buster rm /data/nodes || true
```

So the setup has been moved to compose:
```yaml
setup:
    volumes:
      - "jepsen-shared:/var/jepsen/shared"
    entrypoint: [ 'rm', '-rf', '/var/jepsen/shared/nodes' ]  
```

----

Initial commit didn't include `stop_signal: SIGRTMIN+3` in docker-compose.yml

----

And in some possibly OCD'ish cleanup:

- standardize on `jgoerzen/debian-base-standard:bullseye` for all images

- add `iputils-ping` to control node

- added the few packages that were in `jepsen.os.debian/setup!` but not yet in db node

- make all `apt-get update/install` formatting uniform

----

jepsen-io/etcd and a local test were used for testing on Debian.

### DIFF
--- a/docker/bin/up
+++ b/docker/bin/up
@@ -162,10 +162,6 @@ INFO "Running \`docker compose build\`"
 # shellcheck disable=SC2086
 docker compose --compatibility -p jepsen -f docker-compose.yml ${COMPOSE} ${DEV} build
 
-# We need a fresh share volume each time we start, so we have a correct set of
-# DB hosts. Why does Docker make sharing state SO hard
-docker run --rm -v jepsen_jepsen-shared:/data/ debian:buster rm /data/nodes || true
-
 INFO "Running \`docker compose up\`"
 if [ "${RUN_AS_DAEMON}" -eq 1 ]; then
     # shellcheck disable=SC2086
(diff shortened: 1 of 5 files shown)


aphyr:
Thanks as always!

PR #572, opened 2023-04-17
doc: use built-in parse-long in tutorial.

olepbr:
Hi!

Teeny-tiny PR here.

As noted in 74989931bbdd19d6dcf65e8a9261383d7c57bc21, Clojure 1.11 gave us `parse-long`, ~~which handles `nil` the way we'd like it to, so writing a wrapper for `Long/parseLong` is no longer necessary.~~ so we can use that instead of `Long/parseLong`.

This PR thus amends that section of the tutorial.

Cheers for a great library :–)


olepbr:
Actually, it does not, as it throws `IllegalArgumentException` when it's passed `nil`, my bad! It's still maybe preferable to use it in the wrapper instead of `parseLong`, though?

olepbr:
> It's still maybe preferable to use it in the wrapper instead of `parseLong`, though?

I updated the PR to do this instead. `parse-long-nil` isn't a great name though. Naming things, etc.

aphyr:
okay

PR #576, opened 2023-06-01
Tutorial: fix where parse-long is referenced without being defined

bootjp:
In the original tutorial, parse-long is referenced but not defined.

The definition of parse-long has been added.


aphyr:
`parse-long` is now a part of the Clojure standard library--see https://github.com/jepsen-io/jepsen/pull/572

bootjp:
Sorry, it looks like the version I am using may be out of date.
I will close this pull request.

PR #577, opened 2023-07-22
Fix typo in stolon nemesis.clj

reshke:
Stolon is PostgreSQL HA tool. But `stolon/src/jepsen/stolon/nemesis.clj` says it is going to generate something for MongoDB, which i think is a typo


aphyr:
Yup! Thanks. :-)

PR #580, opened 2023-09-13
kafka: Fix checker worst-realtime-lag default value

enigma:
worst-realtime-lag used to give 0 as default but it's used as a map with :time and :lag keys. This change gives a default with both keys mapping to 0

Fixing that uncovers another problem in `condense-error`, sometimes the error values are not collections, eg with the provided test the error is `{:empty-transaction-graph true}`, and `count` can't be called on that.

See https://github.com/jepsen-io/maelstrom/issues/79 for more information.


aphyr:
Excellent work, thank you!

PR #583, opened 2024-01-26
LXD doc for Debian Bookworm

nurturenature:
Hi,

This PR adds documentation for using the newer LXD tooling that is included with Debian Bookworm.

It's better at setting up and managing it's own networking.
Avoids two horribles and one jfc compared to LXC. :)

Anecdotally, it's a bit faster and more resource efficient. 

It also has support for real VMs that allow using the clock nemesis:

image

The docs for Debian 11 Buster and Ubuntu 14 Trusty were removed for being too old.

### DIFF
--- a/doc/lxc.md
+++ b/doc/lxc.md
@@ -1,5 +1,13 @@
 # How to set up nodes via LXC
 
+#### [Mint 21.2 Victoria](#mint-212-victoria-1)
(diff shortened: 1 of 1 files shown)


aphyr:
Oh nice! I spent a couple days trying to get LXD working in December to no avail--thanks for doing the legwork here.

PR #599, opened 2024-10-21
Fix catch with throw+ instead of throw

wb14123:
The catch is in `try+`, which can catch other things than throwable. So use `throw+` to re-throw the exceptions.


wb14123:
Hey @aphyr does this PR look good to you?

aphyr:
Ah, yes, thank you!

PR #601, opened 2024-10-29
Be able to use other network device in iptables

wb14123:
In `net/iptables`, it uses the network device `eth0` for network operations. But the network device may be different on some machines. So create a function to accept a network device parameter.

### DIFF
--- a/jepsen/src/jepsen/net.clj
+++ b/jepsen/src/jepsen/net.clj
@@ -120,7 +120,7 @@
 (defn- net-shape!
   "Shared convenience call for iptables/ipfilter. Shape the network with tc
   qdisc, netem, and filter(s) so target nodes have given behavior."
-  [_net test targets behavior]
+  [_net test targets behavior dev]
   (let [results (on-nodes test
                           (fn [test node]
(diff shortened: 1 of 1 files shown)


aphyr:
I would say... let's promote this from a reify to a full `(defrecord IPTables [dev ...] ...)`. It'll be more friendly for printing. We can't get away from the static `(def iptables)` without breaking compatibility, but it might make sense to add `(defn iptables* [option-map] ...)` as an underlying constructor--that'll set the groundwork for breaking compat in a later release.

Or, if you just need a one-off, you can reify your own implementation of `net` in your test. I'm for parameterizing the device though! I haven't needed it so far but it makes total sense.

wb14123:
@aphyr is there anything else you want to add to `IPTables` other than `dev`?

aphyr:
No, I think this should basically be good to go. I would like to see this as a defrecord (I can make that change if that's at all tricky for you), but otherwise we're set.

aphyr:
Wow I let this sit FAR too long, I'm sorry!

PR #604, opened 2024-12-26
docker/control installs JDK 21 manually

liyishuai:
The JDK21 hack doesn't work quite well (perhaps due to updates in upstream images).
Not sure if other folks face similar issues, but the changes here have fixed the builds on my side.

### DIFF
--- a/docker/control/Dockerfile
+++ b/docker/control/Dockerfile
@@ -12,16 +12,13 @@ STOPSIGNAL SIGRTMIN+3
 
 ENV LEIN_ROOT true
 
-# JDK21 only in Debian testing
-RUN echo "deb http://deb.debian.org/debian testing main" >> /etc/apt/sources.list
-ADD ./apt-preferences /etc/apt/preferences
-
 #
 # Jepsen dependencies
 #
 RUN apt-get -qy update && \
     apt-get -qy install \
-        curl dos2unix emacs git gnuplot graphviz htop iputils-ping libjna-java openjdk-21-jdk-headless pssh screen vim wget
+        curl dos2unix emacs git gnuplot graphviz htop iputils-ping libjna-java pssh screen vim wget && \
+    curl -L https://download.java.net/java/GA/jdk21.0.2/f2283984656d49d69e91c558476027ac/13/GPL/openjdk-21.0.2_linux-x64_bin.tar.gz | tar -xz --strip-components=1 -C /usr/local/
 
 RUN wget https://raw.githubusercontent.com/technomancy/leiningen/stable/bin/lein && \
     mv lein /usr/bin && \
--- a/docker/control/apt-preferences
+++ b/docker/control/apt-preferences
@@ -1,3 +1,0 @@
-Package: *
-Pin: release a=testing
-Pin-Priority: 499


nurturenature:
Hi Yishuai,

> perhaps due to updates in upstream images

I too noticed this a couple of months ago.
The dependencies for JDK21 in Debian testing have evolved beyond simply grabbing and using the specific Java packages.

This proposed PR is quite heavy handed. It ties Jepsen development and usage to Debian testing, and it completely removes the ssh key setup and sharing between the control and db nodes. IOW, the control node is isolated from the db nodes. Beyond it builds for me, was any testing or usage done?

Would you be open to fixing the PR so that it just replaces:

```Dockerfile
RUN echo "deb http://deb.debian.org/debian testing main" >> /etc/apt/sources.list
ADD ./apt-preferences /etc/apt/preferences
```

with a `wget/curl` of JDK21 from OpenJDK and then installs it?

Note that even in your personal dev environment, if you follow the path of this PR by removing the ` jgoerzen/debian-base`, which neuters the ssh setup scripts, you will also have to recreate all of the same functionality of creating and sharing the ssh keys.

liyishuai:
What architectures do we plan to support? The [JDK archive](https://jdk.java.net/archive/) only provides x86 and arm64. The current version chooses x86. I can add a switch case to support ARM as well.

nurturenature:
Thanks so much for following through!

> The current version chooses x86_64

This is appropriate. Historically the Jepsen docker environment is developed/tested on/with Debian latest/x86_64.

Just one more change is needed, as we're no longer pulling any packages from Debian testing, the `jepsen-control/apt-preferences` file is no longer needed and should be deleted.

I've built images with your changes and run several tests and they work well.

```bash
$ docker exec -it jepsen-control bash
Welcome to Jepsen on Docker
===========================
root@control:/jepsen# java --version
openjdk 21.0.2 2024-01-16
OpenJDK Runtime Environment (build 21.0.2+13-58)
OpenJDK 64-Bit Server VM (build 21.0.2+13-58, mixed mode, sharing)
```

If you'll be running Jepsen tests in GitHub Actions:
  - it's slightly faster and significantly more reliable to publish images to the GitHub Container Registry and pull them in the Action vs building in the Action
  - an example of publishing Jepsen images to the GitHub Container Registry: https://github.com/nurturenature/jepsen-docker/blob/main/.github/workflows/publish-jepsen-images.yml

The free GitHub Actions support reasonable but not large Jepsen tests:
  - some example Jepsen test matrices in GitHub Actions: https://github.com/nurturenature/jepsen-powersync/tree/main/.github/workflows

For local development using Debian 12/Bookworm:
  - Docker Desktop can be a resource hog and limit the # of db nodes you can run
  - Docker CLI, uses different engine, etc, is more efficient and reliable
  - Docker VS Code plugin is a usable GUI if a GUI is desired

nurturenature:
Hi @aphyr ,

This PR has been tested on Debian and Ubuntu, works well, and is ready for merging.

Thanks again Yishuai.

aphyr:
Right, thanks! BTW @nurturenature, if you want to take over maintenance of the Docker setup, I'd be happy to see that spun out into its own repo. :-)

PR #605, opened 2025-01-03
update checker/perf/rate-graph! to behave like checker/perf/point-graph!(latency-raw): elide empty [:f :type] values, use 1s bucket-times

nurturenature:
Hi,

This PR modifies `checker/perf/rate-graph!` to behave like `checker/perf/point-graph!`(latency-raw).

`checker/perf/point-graph!`(latency-raw):
  - elides empty `[:f :type]` values
  - uses 1s bucket-times

image

while the current `checker/perf/rate-graph!`
  - includes empty `[:f :type]` values
  - uses 10s bucket-times

image
 
The inclusion of empty `[:f :type]` values:
  - clutters the graph
  - can invoke a negative initial response, e.g. Oh My God!, there's `:fail`s! `:info`s!
    - squints at graph, are they really 0's? or a need to look at `results.edn`? logs?
 
And the 10s bucket-time can mask a lot of variability in the actual rate, e.g. a challenging nemesis.

New `checker/perf/rate-graph!` with this PR:

image

The updated code for `rate-graph!` follows `point-graph!` 1 for 1.

Thanks!

### DIFF
--- a/jepsen/src/jepsen/checker/perf.clj
+++ b/jepsen/src/jepsen/checker/perf.clj
@@ -572,7 +572,7 @@
   "Writes a plot of operation rate by their completion times."
   [test history {:keys [subdirectory nemeses]}]
   (let [nemeses     (or nemeses (:nemeses (:plot test)))
-        dt          10
+        dt          1
         td          (double (/ dt))
         ; Times might technically be out-of-order (and our tests do this
         ; intentionally, just for convenience)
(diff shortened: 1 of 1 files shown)


aphyr:
I don't think `latency-raw` does *any* bucketing--the idea is it's just the raw points. The graph with bucketing is `latency-quantiles` which IIRC uses the same time buckets as `rate`.

The larger bucket size is intentional--using smaller buckets, as you've shown, tends to make the plot noisier (in the limit, you hit the Nyquist frequency and the plot is nearly useless!). The 1-second buckets you've shown makes it harder to read off approximate request rates, especially at lower throughputs. I'd be open to an adaptive bucket system which scales bucket sizes to request rate, though--up at 1,000 ops/sec a 1-second bucket makes a lot more sense.

I'm up for dropping zero rates! We should only do that when *every* point in the series is zero, and I think that's what the implementation here does, right?

nurturenature:
Thank for the detailed reply.

> The larger bucket size is intentional--using smaller buckets, as you've shown, tends to make the plot noisier ...
> The 1-second buckets you've shown makes it harder to read off approximate request rates, especially at lower throughputs ...

The bucket size has been restored to its original 10s value.

> I'm up for dropping zero rates! We should only do that when every point in the series is zero, and I think that's what the implementation here does, right?

Yes, the `when-let` will only be `nil`, and then `remove`d from the plot, iff the dataset map has no values at all for the given `[:f :type]` map key.

Removing non-existing `:fail`s and `:info`s from the plot was the primary goal.
Changing the bucket size was a misguided attempt at aesthetic improvement.

Thanks again!

aphyr:
Cool beans!

PR #609, opened 2025-02-22
Refactor + fix bugs in `corrupt-file.c`

duckinator:
- Have `main()` always return a value.
- Fix "value too large" error.
- Fix memory leaks.
- Handle MODE_NONE in corrupt(), to simplify `main()`.
- Convert `EXIT_*` to an enum.
- Convert `OPT_*` to an enum.
- Convert `MODE_*` to an enum.
- Avoid unneeded `malloc()`/`free()`.
- Remove always-true conditions.
- Use "typedef struct" to simplify opts type.

### DIFF
--- a/jepsen/resources/corrupt-file.c
+++ b/jepsen/resources/corrupt-file.c
@@ -163,7 +163,7 @@ int validate_opts(struct opts opts) {
     return EXIT_ARGS;
   }
 
-  if ((opts.index < 0) || (opts.mod <= opts.index)) {
+  if (opts.mod <= opts.index) {
     fprintf(stderr, "index %u must fall in [0, %u)\n",
         opts.index, opts.mod);
     return EXIT_ARGS;
--- a/jepsen/resources/corrupt-file.c
+++ b/jepsen/resources/corrupt-file.c
@@ -357,7 +357,7 @@ int corrupt_snapshot(struct opts opts, int fd, off_t file_size, off_t
   for (off_t chunk = opts.index; chunk < chunk_count; chunk += opts.mod) {
     // Where are we corrupting?
     start = chunk_offset(opts, chunk);
-    end   = end + opts.chunk_size;
+    end   = start + opts.chunk_size;
     // Don't read off the end of the region
     if (opts.end < end) {
       end = opts.end;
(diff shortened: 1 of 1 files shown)


jepsen/resources/corrupt-file.c
```diff
@@ -158,7 +163,7 @@ int validate_opts(struct opts opts) {
     return EXIT_ARGS;
   }
 
-  if ((opts.index < 0) || (opts.mod <= opts.index)) {
+  if (opts.mod <= opts.index) {
```
aphyr, line 166: ha HA, living in a world of signed integers strikes again

jepsen/resources/corrupt-file.c
```diff
@@ -703,29 +707,29 @@ int corrupt(struct opts opts) {
 
 /* Deletes the snapshot directory recursively. */
 int clear_snapshots() {
-  size_t limit = sizeof(SNAPSHOT_DIR) + 10;
-  char *cmd = malloc(limit);
-  int written = snprintf(cmd, limit, "rm -rf '%s'", SNAPSHOT_DIR);
+  char cmd[sizeof(SNAPSHOT_DIR) + 10] = {0};
```
aphyr, line 710: Riiiight, because it's fixed size we can use stack allocation here.

jepsen/resources/corrupt-file.c
```diff
@@ -703,29 +707,29 @@ int corrupt(struct opts opts) {
...
 
 /* Go go go! */
 int main (int argc, char **argv) {
   /* Parse args */
-  struct opts opts = default_opts();
+  opts_t opts = default_opts();
   error_t err = argp_parse (&argp, argc, argv, 0, 0, &opts);
   if (err != 0) {
     fprintf(stderr, "Error parsing args: %d\n", err);
+    free(opts.file);
```
aphyr, line 727: Oh good call, I missed this!

aphyr (commented):
Thank you so much Ellen, this is very straightforward! Exactly what I needed. :-)

jepsen/resources/corrupt-file.c
duckinator: `valgrind` is definitely your friend for finding missing `free()` calls.

Issue #612, opened 2025-03-11
Failed to run `lein test` in tutorial due to ssh pam_systemd failures


Wyb1406272043:
### I had an ssh connection error while running the tutorial:
```
ERROR [2025-03-11 14:36:25,635] main - jepsen.cli Oh jeez, I'm sorry, Jepsen broke. Here's why:
clojure.lang.ExceptionInfo: throw+: {:dir "/", :private-key-path nil, :password "root", :username "root", :type :jepsen.control/ssh-failed, :port 22, :strict-host-key-checking false, :host "n1", :sudo nil, :dummy false, :session {:open #object[jepsen.control$session$fn__3147 0x1e74d5d8 "jepsen.control$session$fn__3147@1e74d5d8"], :close #object[jepsen.control$fn__3018$G__2947__3023 0x79f1a2e3 "jepsen.control$fn__3018$G__2947__3023@79f1a2e3"], :log? true, :name [:control "n1"], :lock #object[java.util.concurrent.locks.ReentrantReadWriteLock 0x37b94e5 "java.util.concurrent.locks.ReentrantReadWriteLock@37b94e5[Write locks = 0, Read locks = 0]"], :conn #object[clojure.lang.Atom 0x7e79d881 {:status :ready, :val #jepsen.control.SSHRemote{:session #object[com.jcraft.jsch.Session 0x428cce61 "com.jcraft.jsch.Session@428cce61"]}}]}}
```
<details>

<summary>Detailed information</summary>

```
root@control:/jepsen/jepsen.etcdemo# lein run test
14:36:03.643 [main] INFO jepsen.cli - Test options:
 {:concurrency 5,
 :leave-db-running? false,
 :logging-json? false,
 :ssh
 {:dummy? false,
  :username "root",
  :password "root",
  :strict-host-key-checking false,
  :private-key-path nil},
 :argv ("test"),
 :nodes ["n1" "n2" "n3" "n4" "n5"],
 :test-count 1,
 :time-limit 60}

INFO [2025-03-11 14:36:03,683] jepsen test runner - jepsen.core Command line:
lein run test
INFO [2025-03-11 14:36:03,711] jepsen test runner - jepsen.core Running test:
{:remote #jepsen.control.SSHRemote{:session nil}
 :concurrency 5
 :db
 #object[jepsen.etcdemo$db$reify__331
         "0x4331be30"
         "jepsen.etcdemo$db$reify__331@4331be30"]
 :leave-db-running? false
[...]

PR #614, opened 2025-03-11
Close the UsePAM service

Wyb1406272043:
The UsePAM service was turned off to prevent ssh connection timeout errors

### DIFF
--- a/docker/node/Dockerfile
+++ b/docker/node/Dockerfile
@@ -29,6 +29,7 @@ RUN chmod +x /usr/local/preinit/03-setup-jepsen
 
 # Configure SSHD
 RUN sed -i "s/#PermitRootLogin prohibit-password/PermitRootLogin yes/g" /etc/ssh/sshd_config
+RUN sed -i 's/^#*UsePAM.*/UsePAM no/' /etc/ssh/sshd_config
 
 # Enable SSH server
 ENV DEBBASE_SSH enabled


aphyr:
@nurturenature @jackyzha0, just wanted to get your read on this--I know you two are responsible for a lot of the Docker setup. @Wyb1406272043 reports that they've been seeing PAM-related timeouts, but I'm a little wary of disabling PAM support in the SSH daemon--I know that's involved in session management stuff as well as just auth.

nurturenature:
Hi,

My goal has been to keep the Docker setup working with Debian latest
- with as minimal Docker and Debian mods as possible
- primarily for use in a CI, e,g, GitHub Actions, environment
- secondarily for systems that have complex Docker Compose environments that are too hard to mimic with LXC/LXD

I can confirm that the current repository, without this PR, works fine
- using Debian latest, Bookworm, as a host
- using Ubuntu latest, 24.04, as a host with GitHub Actions
  - GitHub Actions was using 22.04 until several months ago and that was also consistently working

Looking at #612 

> I tried the ubuntu22.04 virtual machine and container running the official docker image provided by jepsen, with the same result, and ssh n1 took 25 seconds to connect.

Could this be a container in a container security issue?
Docker needs special handling when not running on an actual host.

I recommend not applying this PR
- not needed for Debian, Jepsen's "home" environment
- disabling PAM when it's not the root cause seems heavy handed
  > Changing UsePAM in /etc/ssh/sshd_config to no resolves this issue, but it seems that the root cause is that systemd-logind service is somehow not starting successfully,I don't know what caused it.
- doesn't feel like the issue is well understood

@Wyb1406272043, you could try the Docker images at https://github.com/nurturenature/jepsen-docker

These are the images I build into the GitHub Container Registry for faster more reliable use of Jepsen in GitHub Actions and are known to work with  Ubuntu latest.

nurturenature:
So... I spoke too soon. 

Trying on a fresh OS/Docker/Jepsen install:
- there have been multiple systemd changes that require some type of credentials
- not all modules, apps, configs, have caught up, are configured
- so there are actually multiple systemd failures in the syslog when running with Docker
  -  systemd-login is what was initially noticed, but there's several more

We've really tried to avoid running privileged Docker containers, but is that reasonably possible now?!?

@Wyb1406272043, if you edit `jepsen/docker/template/docker-compose.yml`, and add a `privileged: true` entry:
```yaml
x-node:
  &default-node
  privileged: true
  build: ./node
...
```

Docker will boot the node containers without error and will be accessible from the control node.
```bash
$ ./bin/up

# from another terminal
$ ./bin/console 
Welcome to Jepsen on Docker
...
# ssh login is now fast and error free
root@control:/jepsen# ssh n1
Linux n1 6.1.0-31-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.128-1 (2025-02-07) x86_64
...
```

This is actually a real bug in the current Jepsen Docker environment.
I'll do some research and propose a fix.

Apologies for my initial "it's working for me" assumptions.

Wyb1406272043:
@nurturenature Ok, I am just a beginner and not familiar with ssh and containers, I had the same concern when submitting this pr, thanks for your answer and solution, I will try.

PR #634, opened 2026-04-24
jepsen.control.util/ls-full: fix typo, mark deprecated, and change uses of ls-full to ls

nurturenature:
Hi,

Using the `bitflip` Nemesis exposed a bug that was introduced in [Commit 0479c46](https://github.com/nurturenature/jepsen/commit/0479c464227afa70378e392681e879e536a18c04).

This PR fixes the typo in `jepsen.control.util/ls-full`.

And as the original `ls-full` docstring said "TODO: deprecate this in favor of ls {:full-path? true}"

- marks `ls-full` as `^:deprecated`
- changes the two uses of `ls-full` in Jepsen to `ls`

Also, I think it reads more closely to the Nemesis' intent, corrupt a random file from the given directory

```clj
(rand/nth (cu/ls file {:recursive? true :types [:file] :full-path? true}))
```

Thanks!

### DIFF
--- a/jepsen/src/jepsen/control/util.clj
+++ b/jepsen/src/jepsen/control/util.clj
@@ -96,13 +96,13 @@
      ; Find's traversal order is by inode structure, so we sort for stability
      (sort paths))))
 
-(defn ls-full
-  "Like ls, but prepends dir to each entry. TODO: deprecate this in favor of ls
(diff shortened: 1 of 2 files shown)


aphyr:
Oof, thank you!

